PR #36, opened 2015-07-22
Help messages

gkiki90:


cli/commands.go
```diff
@@ -7,7 +7,7 @@ var (
 		{
 			Name:      "setup",
 			ShortName: "s",
-			Usage:     "",
+			Usage:     "Setup initialize the specified collection, it's required befor using a collection.",
```
viktorbenei: ~~Setup~~ Initialize the specified collection, it's required befor`e` using a collection.

cli/commands.go
```diff
@@ -16,7 +16,7 @@ var (
 		{
 			Name:      "update",
 			ShortName: "u",
-			Usage:     "",
+			Usage:     "Updates the collection, if no --collection flag provided, all collections will updated.",
```
viktorbenei: Update~~s~~ the collection. If no --collection flag provided, all collections will updated.

cli/commands.go
```diff
@@ -25,7 +25,7 @@ var (
 		{
 			Name:      "download",
 			ShortName: "d",
-			Usage:     "",
+			Usage:     "Downloads the step wit provided --id, and --version, from specified --collection into local step cache. If no --version defined, the latest version of step will be cached.",
```
viktorbenei: Download the step with provided --id and --version, from specified --collection, into local step downloads cache. If no --version defined, the latest version of the step (latest found in the collection) will be downloaded into the cache.

cli/commands.go
```diff
@@ -36,7 +36,7 @@ var (
 		{
 			Name:      "activate",
 			ShortName: "a",
-			Usage:     "",
+			Usage:     "Copies the step with specified --id, and --version, into provided path. If --version flag is unset, the latest version of step will be used. If --copyyml flag is set, step.yml will copied into given path.",
```
viktorbenei: Copy the step with specified --id, and --version, into provided path. If --version flag is not set, the latest version of the step will be used. If --copyyml flag is set, step.yml will be copied to the given path.

cli/flags.go
```diff
@@ -84,7 +84,7 @@ var (
 	flCopyYML = cli.StringFlag{
 		Name:  CopyYMLKey + ", " + copyYMLKeyShort,
 		Value: "",
-		Usage: "",
+		Usage: "Path where the step.yml will copied.",
```
viktorbenei: Path where the selected/activated step's step.yml will be copied.

PR #54, opened 2015-07-29
New envman models

gkiki90:


models/models_methods.go
```diff
@@ -7,42 +7,11 @@ import (
...
-	if err != nil {
-		return err
-	}
-	(*env)[optionsKey] = opts
-	return nil
-}
-
 // Validate ...
-func (env EnvironmentItemModel) Validate() error {
+func Validate(env envmanModels.EnvironmentItemModel) error {
```
viktorbenei: instead of just `Validate` it should be `ValidateEnvironmentItemModel`

PR #68, opened 2015-08-07
source increased with commit hash

gkiki90:


models/models.go
```diff
@@ -6,7 +6,8 @@ import (
 
 // StepSourceModel ...
 type StepSourceModel struct {
-	Git *string `json:"git,omitempty" yaml:"git,omitempty"`
+	Git    *string `json:"git,omitempty" yaml:"git,omitempty"`
```
viktorbenei: these should not be pointers, simple `omitempty` should be enough

stepman/git.go
```diff
@@ -24,18 +26,42 @@ func DoGitClone(uri, pth string) error {
...
-	if version == "" {
-		return errors.New("Git Clone 'version' missing")
+	if commithash == "" {
+		return errors.New("Git Clone 'hash' missing")
+	}
+	if err := RunCommand("git", []string{"clone", "--recursive", uri, pth, "--branch", version}...); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("git", []string{"rev-parse", "HEAD"}...)
```
viktorbenei: move this into a separate `util` function

stepman/git.go
```diff
@@ -24,18 +26,42 @@ func DoGitClone(uri, pth string) error {
...
 		return errors.New("Git Clone 'pth' missing")
 	}
-	if version == "" {
-		return errors.New("Git Clone 'version' missing")
+	if commithash == "" {
+		return errors.New("Git Clone 'hash' missing")
+	}
+	if err := RunCommand("git", []string{"clone", "--recursive", uri, pth, "--branch", version}...); err != nil {
+		return err
+	}
```
viktorbenei: cleanup (remove the git dir) if anything else after this fails - ex if the `hash` check fails. Do it with a `defer` block?

PR #77, opened 2015-08-08
run CI on _step_template

viktorbenei:


viktorbenei:
Requires bitrise v0.9.8, where 'source' is not required anymore in validation, only if a step is used through `stepman` (stepman validates it, for steplib items)

PR #204, opened 2016-08-10
Log - path ref fix

viktorbenei:
I guess there's no point in printing `YMLPath` in the `else` case, when we know that it's empty ;)


godrei:
@viktorbenei  you're right

PR #219, opened 2016-11-03
prepare for deprecating step versions

godrei:
* replaced StepGroupInfoModel with DeprecationInfoModel and AssetURLMap
* added Deprecation to step model
* removed Source property from StepInfoModel
* test updates
* utility revision
* version bump to 0.9.26-pre

### DIFF
--- a/cli/export.go
+++ b/cli/export.go
@@ -54,16 +54,15 @@ func convertToLatestSpec(stepLib models.StepCollectionModel) models.StepCollecti
 
 	latestSteps := models.StepHash{}
 	for stepID, stepGroup := range steps {
-		groupInfo := stepGroup.Info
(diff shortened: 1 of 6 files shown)


viktorbenei:
Rather do this:

- keep deprecation info in the `step-info.yml`, as it is right now
- just add a `deprecated_versions` array
- ideally the deprecated versions array could include version ranges, like `0.0.1 - 1.0.0`, or expressions like `<= 1.5.0`

PRO:

- deprecation infos at one place, not spread through step.yml s
- easier to manage, easier to check
- does not require polluting `step.yml`, nor its dir with yet another yml file

CON:

- you can't specify different deprecation messages for different versions; but I don't think this would be an issue
- anything else?

PR #239, opened 2017-11-28
Slim steplib

kdobmayer:


stepman/util.go
```diff
@@ -279,6 +279,29 @@ func generateStepLib(route SteplibRoute, templateCollection models.StepCollectio
...
+		SteplibSource:         collection.SteplibSource,
+		DownloadLocations:     collection.DownloadLocations,
+		AssetsDownloadBaseURI: collection.AssetsDownloadBaseURI,
+	}
+	steps := models.StepHash{}
+
+	for stepID, stepGroupModel := range collection.Steps {
+		steps[stepID] = models.StepGroupModel{
+			Info:     stepGroupModel.Info,
+			Versions: map[string]models.StepModel{stepGroupModel.LatestVersionNumber: stepGroupModel.Versions[stepGroupModel.LatestVersionNumber]},
```
godrei: LatestVersion property is missing, it is used to find the desired key in Versions map.

PR #262, opened 2019-01-08
validate tag format adheres to semver

lszucs:


cli/share_create.go
```diff
@@ -45,8 +66,8 @@ func create(c *cli.Context) error {
 
 	// Input validation
 	tag := c.String(TagKey)
-	if tag == "" {
-		fail("No Step tag specified")
+	if err := validateTag(tag); err != nil {
+		fail("validate tag: %s", err)
```
trapacska: `validate tag: invalid semver format %s; valid format is 'x.y.z'` is not informative enough. Please change validate tag to something like issue with step tag, and also include in the error message that what is the actual one

PR #268, opened 2019-03-27
Adding meta field to step model.

lpusok:


_tests/integration/step_info_test.go
```diff
@@ -158,7 +158,7 @@ inputs:
...
...:["osx-10.10"],"project_type_tags":["ios","android","xamarin"],"type_tags":["script"],"deps":{"brew":[{"name":"git"},{"name":"wget"}],"apt_get":[{"name":"git"},{"name":"wget"}]},"is_requires_admin_user":true,"is_always_run":false,"is_skippable":false,"run_if":"","timeout":0,"meta":{"bitrise.io.addons.required":[{"addon_id":""}],"bitrise.io.addons.optional":[{"addon_id":""}]},"inputs":[{"example_step_input":"Default Value - you can leave this empty if you want to","opts":{"is_expand":true,"skip_if_empty":false,"title":"Example Step Input","description":"Description of this input.\n\nCan be Markdown formatted text.\n","summary":"Summary. No more than 2-3 sentences.","category":"","is_required":true,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false}}],"outputs":[{"EXAMPLE_STEP_OUTPUT":null,"opts":{"is_expand":true,"skip_if_empty":false,"title":"Example Step Output","description":"Description of this output.\n\nCan be Markdown formatted text.\n","summary":"Summary. No more than 2-3 sentences.","category":"","is_required":false,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false}}]},"definition_pth":"test-step/step.yml"}`
```
trapacska: the values for these meta fields are missing: `"meta":{"bitrise.io.addons.required":[{"addon_id":""}],"bitrise.io.addons.optional":[{"addon_id":""}]}`

models/models.go
```diff
@@ -69,6 +69,16 @@ type StepToolkitModel struct {
 	Go   *GoStepToolkitModel   `json:"go,omitempty" yaml:"go,omitempty"`
 }
 
+// metaModel contains arrays with addons' identifiers that are required or optional to be connected for the step to function correctly
+type metaModel struct {
```
trapacska: I think this model should be simply `map[string]interface{}` so it would be an universal one, so we won't need to modify this field each time a new meta info need to be stored

_tests/integration/step_info_test.go
```diff
@@ -195,6 +195,9 @@ type_tags:
 is_requires_admin_user: true
 is_always_run: false
 is_skippable: false
+meta:
+  bitrise.io.addons.required: [{addon_id:"addons-testing"}]
```
godrei: This is not a valid YML, here you defined the `bitrise.io.addons` with JSON values, which is fine, but `[{addon_id:"addons-testing"}]` is not valid JSON, that's why the key is missing spotted by Tomi: https://github.com/bitrise-io/stepman/pull/268/files#r270055538

`[{addon_id:"addons-testing"}]` -> ``[{"addon_id":"addons-testing"}]``

Other than that I would recommend specifying a bit more complex meta to tests if further options will be kept and parsed properly:

```
meta:
  bitrise.io.addons.required: 
    - addon_id: "addons-testing"
      addon_params: "--token TOKEN"
      addon_options: 
        required: true
        title: "Testing Addon"
    - addon_id: "addons-ship"
      addon_params: "--token TOKEN"
      addon_options: 
        required: true
        title: "Ship Addon"
  bitrise.io.addons.optional: [{"addon_id":"addons-testing"}] # json value
```

models/models.go
godrei: agree we will not use this property of steps, we just need to keep it when we are parsing step and bitrise ymls.

PR #269, opened 2019-04-08
required addons

lpusok:
Previously: https://github.com/bitrise-io/stepman/pull/268


models/models.go
```diff
@@ -69,6 +69,10 @@ type StepToolkitModel struct {
 	Go   *GoStepToolkitModel   `json:"go,omitempty" yaml:"go,omitempty"`
 }
 
+type addonReference struct {
```
godrei: is it used?

lpusok: good catch.

models/parse_util.go
```diff
@@ -0,0 +1,62 @@
+package models
+
+import (
+	"fmt"
+)
+
+func normalizeMapStringInterface(source map[string]interface{}) (map[string]interface{}, error) {
```
godrei: NOTE<not sure>: what about: `jsonMarshallable/jsonify` the other could be `recursiveJSONMarshallable/JSONify`, since the point of these methods is to make the type json compatible/marshallable

models/models_methods.go
```diff
@@ -148,7 +156,7 @@ func (step StepModel) ValidateInputAndOutputEnvs(checkRequiredFields bool) error
 }
 
 // AuditBeforeShare ...
-func (step StepModel) AuditBeforeShare() error {
+func (step *StepModel) AuditBeforeShare() error {
```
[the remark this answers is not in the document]
lpusok: It is prioritizing consistency in this case.

Gopkg.lock
```diff
@@ -55,6 +47,14 @@
   pruneopts = "UT"
   revision = "f451b19ba4d087d8272d411ad91af5f4218fd517"
 
+[[projects]]
+  branch = "master"
+  digest = "1:eda2296387e3dae410b4ceb25e4db754d6bb552666584ef93ca1d98aa12920df"
+  name = "github.com/bitrise-tools/colorstring"
```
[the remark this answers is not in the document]
lpusok: It seems to have have been resolved through the master branch, no references found otherwise.

PR #283, opened 2020-01-20
Activate exported function

godrei:


cli/commands.go
```diff
@@ -66,19 +66,7 @@ var (
...
 		},
-		{
-			Name:   "activate",
-			Usage:  "Copy the step with specified --id, and --version, into provided path. If --version flag is not set, the latest version of the step will be used. If --copyyml flag is set, step.yml will be copied to the given path.",
-			Action: activate,
-			Flags: []cli.Flag{
-				flCollection,
-				flID,
-				flVersion,
-				flPath,
```
lpusok: flPath and flCopyYML does not seem to be used elsewhere, so the original definiton can be removed.

cli/activate.go
```diff
@@ -1,138 +1,188 @@
...
+			Name:  VersionKey + ", " + versionKeyShort,
+			Usage: "Step version.",
+		},
+		cli.StringFlag{
+			Name:  PathKey + ", " + pathKeyShort,
+			Usage: "Path where the step will copied.",
+		},
+		cli.StringFlag{
+			Name:  CopyYMLKey + ", " + copyYMLKeyShort,
+			Usage: "Path where the selected/activated step's step.yml will be copied.",
```
lpusok: It can be just activated instead of selected/activated.

cli/commands.go
lpusok: NOTE: Unrelated flShort and flStepYML are also unused.

cli/activate.go
```diff
@@ -1,138 +1,188 @@
...
+		return nil
+	},
+}
+
 func activate(c *cli.Context) error {
-	// Input validation
 	collectionURI := c.String(CollectionKey)
 	if collectionURI == "" {
-		log.Fatalf("No step collection specified")
+		return fmt.Errorf("activate: no steplib specified")
```
lpusok: Check error strings format as discussed.

cli/activate.go
```diff
@@ -1,138 +1,188 @@
...
+		return models.StepModel{}, "", fmt.Errorf("query step: %s steplib does not contain %s step", library.SteplibSource, id)
+	}
+	if !versionFound {
+		return models.StepModel{}, "", fmt.Errorf("query step: %s steplib does not contain %s step %s version", library.SteplibSource, id, version)
 	}
 
-	// If version doesn't provided use latest
 	if version == "" {
-		latest, err := collection.GetLatestStepVersion(id)
+		latest, err := library.GetLatestStepVersion(id)
```
lpusok: This seems to related to the removed ` Check step exist in local cache` part. It may be removed as the GetStep function also handles version is empty string case.

cli/activate.go
```diff
@@ -1,138 +1,188 @@
...
-		}
+	if exist, err := pathutil.IsPathExists(dest); err != nil {
+		return fmt.Errorf("copy step.yml: failed to check if %s path exist: %s", dest, err)
+	} else if exist {
+		return fmt.Errorf("copy step.yml: %s already exist", dest)
 	}
 
+	stepCollectionDir := stepman.GetStepCollectionDirPath(route, id, version)
+	stepYMLSrc := stepCollectionDir + "/step.yml"
+	if err := command.CopyFile(stepYMLSrc, dest); err != nil {
```
lpusok: Could use filepath.Join here: `stepYMLSrc := stepCollectionDir + "/step.yml"`

cli/activate.go
godrei: later we need the exact version: 
- https://github.com/bitrise-io/stepman/pull/283/files#diff-6e1b0a2810695a1f8a5122e23f318696R88
- https://github.com/bitrise-io/stepman/pull/283/files#diff-6e1b0a2810695a1f8a5122e23f318696R98

PR #288, opened 2020-01-29
Tool 1606 add maintainer field

trapacska:


models/models_methods.go
```diff
@@ -28,12 +28,18 @@ const (
 func (stepInfo StepInfoModel) String() string {
```
trapacska: Screenshot 2020-01-29 at 11 34 17

cli/step_info.go
```diff
@@ -210,12 +210,20 @@ func QueryStepInfoFromLibrary(library, id, version string) (models.StepInfoModel
 	stepDir := stepman.GetStepCollectionDirPath(route, id, stepVersion.Version)
 	stepDefinitionPth := filepath.Join(stepDir, "step.yml")
 
+	infoPath := stepman.GetStepGlobalInfoPath(route, id)
+
+	groupInfo, _, err := stepman.ParseStepGroupInfoModel(infoPath)
```
lpusok: Think should add a seperate validation that ensures that step-info.yml is valid for all of the steplib steps, or have a fallback in the case failed to parse GroupInfo. As currently QueryStepInfoFromLibrary is called on every step activate, this seems warranted.

PR #292, opened 2020-03-24
Fix local StepLib step activation issue

godrei:


_tests/integration/activate_test.go
```diff
@@ -0,0 +1,89 @@
...
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, restoreFunc())
+	}()
+
+	// setup a local collection
+	stepLibDir, err := cloneDeafultStepLib()
+	require.NoError(t, err)
+
+	out, err := command.New(binPath(), "setup", "-c", "file://"+stepLibDir).RunAndReturnTrimmedCombinedOutput()
```
lpusok: I think there should be a better way to run these tests, without needing to precompile the binary.

PR #295, opened 2020-09-23
[PLAT-92] Generate step-info.yml for new steps

zsolt-marta-bitrise:
## Context

Jira: https://bitrise.atlassian.net/browse/PLAT-92

We want to generate a `step-info.yml` automatically during sharing to avoid issues and manual work needed.

### General

- Renamed some variables/methods in `util.go` to better convey their purpose
- Added tests for `paths.go`

### In create stage

- Fixed grammatical and phrasing issues in logs / status reports
- New method for getting the `step-info.yml` path: `GetStepGlobalInfoPath`
- Check if the step is new by checking the step root (a.k.a. step group) directory
- Create the default `step-info.yml` if the step is new

### In finish stage
- When checking the git status, we should use `-u` to list all changes
- If the `step-info.yml` is added, add it to the commit as well

### DIFF
--- a/cli/share_create.go
+++ b/cli/share_create.go
@@ -54,7 +54,7 @@ func validateTag(tag string) error {
 	return nil
 }
 
-func createDefaultStepGroupSpec(stepDirInSteplib string) error {
+func createDefaultStepGroupSpec(route stepman.SteplibRoute, id string) error {
 	model := models.StepGroupInfoModel{}
 	model.Maintainer = "community"
 
@@ -63,7 +63,7 @@ func createDefaultStepGroupSpec(stepDirInSteplib string) error {
 		return err
 	}
 
-	pth := path.Join(stepDirInSteplib, "step-info.yml")
+	pth := stepman.GetStepGlobalInfoPath(route, id)
 	return fileutil.WriteBytesToFile(pth, marshalled)
 }
 
@@ -128,11 +128,6 @@ func create(c *cli.Context) error {
 	collection, err := stepman.ParseStepCollection(collectionSpecPath)
 	if err != nil {
 		fail("Failed to read step spec, error: %s", err)
-	}
-	if !collection.IsAnyStepVersionExist(stepID) {
-		if err := createDefaultStepGroupSpec(stepDirInSteplib); err != nil {
-			fail("Failed to create step group spec for new step: %s", err)
-		}
 	}
 
 	// Clone Step to tmp dir
@@ -244,6 +239,12 @@ func create(c *cli.Context) error {
 	}
 	if err := fileutil.WriteBytesToFile(stepYMLPathInSteplib, stepBytes); err != nil {
 		fail("Failed to write Step to file, err: %s", err)
+	}
+
+	if !collection.IsAnyStepVersionExist(stepID) {
+		if err := createDefaultStepGroupSpec(route, stepID); err != nil {
+			fail("Failed to create step group spec for new step: %s", err)
+		}
 	}
 
 	log.Printf("your Step (%s@%s) has been added to the local StepLib (%s).", share.StepID, share.StepTag, stepDirInSteplib)
--- a/cli/share_create.go
+++ b/cli/share_create.go
@@ -76,7 +76,7 @@ func create(c *cli.Context) error {
 		fail("No Step url specified")
 	}
 
-	stepID := c.String(StepIDKEy)
+	stepID := c.String(StepIDKey)
 	if stepID == "" {
 		stepID = getStepIDFromGit(gitURI)
 	}
@@ -85,7 +85,7 @@ func create(c *cli.Context) error {
 	}
 	r := regexp.MustCompile(`[a-z0-9-]+`)
 	if find := r.FindString(stepID); find != stepID {
-		fail("StepID doesn't conforms to: [a-z0-9-]")
+		fail("StepID doesn't conform to: [a-z0-9-]+")
 	}
 
 	route, found := stepman.ReadRoute(share.Collection)
@@ -97,14 +97,14 @@ func create(c *cli.Context) error {
 	if exist, err := pathutil.IsPathExists(stepYMLPathInSteplib); err != nil {
 		fail("Failed to check step.yml path in steplib, err: %s", err)
 	} else if exist {
-		log.Printf("Step already exist in path: %s", stepDirInSteplib)
-		log.Warnf("For sharing it's required to work with a clean Step repository.")
-		if val, err := goinp.AskForBool("Would you like to overwrite local version of Step?"); err != nil {
+		log.Printf("Step already exists in path: %s", stepDirInSteplib)
+		log.Warnf("Sharing requires to work in a clean Step repository.")
+		if val, err := goinp.AskForBool("Would you like to overwrite the local version of the Step?"); err != nil {
 			fail("Failed to get bool, err: %s", err)
 		} else {
 			if !val {
-				log.Errorf("Unfortunately we can't continue with sharing without an overwrite exist step.yml.")
-				fail("Please finish your changes, run this command again and allow it to overwrite the exist step.yml!")
+				log.Errorf("Unfortunately, we can't continue with sharing without overwriting the existing step.yml.")
+				fail("Please finish your changes, run this command again, and allow it to overwrite the existing step.yml!")
 			}
 		}
 	}
@@ -141,7 +141,7 @@ func create(c *cli.Context) error {
 	}
 	var stepModel models.StepModel
 	if err := yaml.Unmarshal(bytes, &stepModel); err != nil {
-		fail("Failed to unmarchal Step, err: %s", err)
+		fail("Failed to unmarshal Step, err: %s", err)
 	}
 
 	commit, err := repo.RevParse("HEAD").RunAndReturnTrimmedCombinedOutput()
@@ -171,7 +171,7 @@ func create(c *cli.Context) error {
 		}
 
 		if len(options.ValueOptions) > 0 && value == "" {
-			log.Warnf("Step input with 'value_options', should contain default value!")
+			log.Warnf("Step input with 'value_options' should contain a default value!")
 			fail("Missing default value for Step input (%s).", key)
 		}
 	}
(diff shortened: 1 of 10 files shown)


stepman/util.go
```diff
@@ -194,7 +194,7 @@ func addStepVersionToStepGroup(step models.StepModel, version string, stepGroup
 	return stepGroup, nil
 }
 
-func generateStepLib(route SteplibRoute, templateCollection models.StepCollectionModel) (models.StepCollectionModel, error) {
+func generateStepLibCollection(route SteplibRoute, templateCollection models.StepCollectionModel) (models.StepCollectionModel, error) {
```
trapacska, line 197: I would go with the hybrid `generateStepCollection`

stepman/util.go
```diff
@@ -301,7 +301,7 @@ func generateStepLib(route SteplibRoute, templateCollection models.StepCollectio
 	return collection, nil
 }
 
-func generateSlimStepLib(collection models.StepCollectionModel) models.StepCollectionModel {
+func generateSlimStepLibModel(collection models.StepCollectionModel) models.StepCollectionModel {
```
trapacska, line 304: same as above

trapacska (changes requested):

trapacska (approved):

cli/share_create.go
```diff
@@ -54,6 +54,31 @@ func validateTag(tag string) error {
 	return nil
 }
 
+func getDefaultStepGroupSpec() models.StepGroupInfoModel {
+	model := models.StepGroupInfoModel{}
```
trapacska, line 58:
```
return models.StepGroupInfoModel{
 Maintainer: "community",
}
```

cli/share_finish.go
```diff
@@ -43,23 +46,35 @@ func finish(c *cli.Context) error {
...
+	if gitstatus == "" {
 		log.Warnf("No git changes, it seems you already called this command")
 		fmt.Println()
 		printFinishShare()
 		return nil
 	}
 
 	stepDirInSteplib := stepman.GetStepCollectionDirPath(route, share.StepID, share.StepTag)
-	stepYMLPathInSteplib := stepDirInSteplib + "/step.yml"
+	stepYMLPathInSteplib := path.Join(stepDirInSteplib, "step.yml")
```
trapacska, line 61: NOTE: we ususally use `filepath` instead of `path`

cli/share_finish.go
```diff
@@ -43,23 +46,35 @@ func finish(c *cli.Context) error {
...
 	log.Printf("new step.yml: %s", stepYMLPathInSteplib)
 	if err := repo.Add(stepYMLPathInSteplib).Run(); err != nil {
 		fail(err.Error())
 	}
+	// add auto generated step-info.yml for new steps
+	stepInfoYMLPathInSteplib := stepman.GetStepGlobalInfoPath(route, share.StepID)
+	if exists, err := pathutil.IsPathExists(stepInfoYMLPathInSteplib); err == nil {
+		if exists && strings.Contains(gitstatus, path.Base(stepInfoYMLPathInSteplib)) {
```
trapacska, line 69: It might be more understandable as extracted to a new fn

cli/share_finish.go
```diff
@@ -43,23 +46,35 @@ func finish(c *cli.Context) error {
...
+			log.Printf("new step-info.yml: %s", stepInfoYMLPathInSteplib)
+			if err := repo.Add(stepInfoYMLPathInSteplib).Run(); err != nil {
+				fail(err.Error())
+			}
+		}
+	} else {
+		fail(err.Error())
```
trapacska, line 76: not sure if we could extend the error message a little

trapacska (changes requested):

cli/share_create_test.go
```diff
@@ -15,9 +20,19 @@ func TestValidateTag(t *testing.T) {
...
+	// Given
+	expected := models.StepGroupInfoModel{
+		Maintainer: "community",
 	}
+
+	// When
+	actual := getDefaultStepGroupSpec()
+
+	// Then
+	assert.Equal(t, expected, actual)
```
adborbas, line 37: I would suggest using `require.*` instead of `assert.*`. The difference is, `require` fails if the condition is not met, while `assert` only returns a bool. https://github.com/stretchr/testify#require-package

stepman/paths.go
```diff
@@ -200,6 +200,12 @@ func GetLibraryBaseDirPath(route SteplibRoute) string {
 	return path.Join(GetCollectionsDirPath(), route.FolderAlias, "collection")
 }
 
+// GetStepCollectionSpecPath ...
+// Location of steplib.yml of the collection marked by this route
```
adborbas, line 204: I guess you wanted to write these in a single sentence.

stepman/paths_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"os"
+	"path"
+	"testing"
+)
+
+const (
+	GivenSteplibURI  = "https://github.com/bitrise-io/steplib"
```
adborbas, line 12: It does not really make any difference in the case of a `_test.go` file, but I would rather keep constants and `GivenRoute ` camlCase. WDYT?

stepman/paths_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func GivenRoute() SteplibRoute {
+	return SteplibRoute{
+		SteplibURI:  GivenSteplibURI,
+		FolderAlias: GivenFolderAlias,
+	}
+}
+
+func Test_GivenHomeDir_WhenGetStepmanDirPathCalled_ThenGoodPathReturned(t *testing.T) {
```
adborbas, line 26: 👍 for using descriptive names and // Given ....

stepman/paths_test.go
```diff
@@ -0,0 +1,89 @@
...
+func Test_GivenRoute_WhenGetLibraryBaseDirPathCalled_ThenGoodPathReturned(t *testing.T) {
+	// Given
+	route := GivenRoute()
+	expected := path.Join(GetCollectionsDirPath(), route.FolderAlias, "collection")
+
+	// When
+	actual := GetLibraryBaseDirPath(route)
+
+	// Then
+	assert.Equal(t, expected, actual)
```
adborbas, line 61: 🙈 Please be consistent with the order of `expected` and `actual`. One function above it has a different order then here. Sorry, just nitpicking.

stepman/paths_test.go
```diff
@@ -0,0 +1,89 @@
...
+func GivenRoute() SteplibRoute {
+	return SteplibRoute{
+		SteplibURI:  GivenSteplibURI,
+		FolderAlias: GivenFolderAlias,
+	}
+}
+
+func Test_GivenHomeDir_WhenGetStepmanDirPathCalled_ThenGoodPathReturned(t *testing.T) {
+	// Given
+	err := os.Setenv("HOME", GivenHomePath)
```
adborbas, line 28: Just thinking, wouldn't it mess up the `HOME` env running this test?

stepman/paths.go
zsolt-marta-bitrise: It's interesting, all the other methods have this format 🤔

cli/share_create_test.go
zsolt-marta-bitrise: And so `assert` will not even fail the test? 😮

stepman/paths.go
adborbas: I think they are messed up as well :D We usually use `<functionname> ...` to get pass golint. But you can give proper comments to exposed functions.

stepman/paths_test.go
zsolt-marta-bitrise: Do you mean to start with a lowercase here? like `GivenSteplibURI` ?

adborbas: yep

cli/share_create_test.go
adborbas: This is my understanding based on the documentation.

adborbas: I did not check it though ...

stepman/paths_test.go
zsolt-marta-bitrise: @trapacska wdyt? Not sure how we could test this

zsolt-marta-bitrise: Ok I added logic to reset the HOME variable after the test

cli/share_create_test.go
zsolt-marta-bitrise: I just tried it, and it fails the test.
I would go with `assert`, because that naming is widely adopted already in testing contexts, and `require` is usually used for input argument validation. 
The additional feature of stopping the test is not required because the assertions are independent, and for example in `Test_GivenHomeDir_WhenGetStepmanDirPathCalled_ThenGoodPathReturned` I use `require` because if the setup fails, that should indeed stop the test.

So in general I'd vote for `assert`

stepman/paths_test.go
zsolt-marta-bitrise: Nice catch! I'm always confused, Go does a lot of things the way around of what I'm used to 🙈

stepman/paths_test.go
zsolt-marta-bitrise: Btw, we've looked into this and it will not overwrite the variable as this change is visible only in this process. But, nevertheless, it's a good pattern to clean up after each test, so I'll leave it.

cli/share_create_test.go
adborbas: assert is fine

cli/share_create.go
```diff
@@ -171,22 +198,27 @@ func create(c *cli.Context) error {
...
-		log.Warnf("Step summary should contains maximum (%d) characters, actual: (%d)!", maxSummaryLength, utf8.RuneCountInString(*stepModel.Summary))
+		log.Warnf("Step summary should contain maximum (%d) characters, actual: (%d)!", maxSummaryLength, utf8.RuneCountInString(*stepModel.Summary))
 	}
 	log.Donef("step is valid")
 
 	// Copy step.yml to steplib
 	fmt.Println()
 	log.Infof("Integrating the Step into the Steplib...")
 
+	isStepNew, err := isStepNew(route, stepID)
```
adborbas, line 217: Quick question, why do we call `isStepNew` here if we only use the value at line 256?

zsolt-marta-bitrise: Because by that time we've created the directory and we couldn't detect if the step was new (because the check will always return false)

adborbas (approved):

trapacska (approved):

PR #309, opened 2022-08-04
Add no output timeout property

lpusok:
#### Context

Added a new Step property `no_output_timeout`, that can be used to set the timeout (in seconds) until the Step is aborted when no stdout/stderr is received.

This Step level property will override any global value set via `BITRISE_NO_OUTPUT_TIMEOUT` env var.

#### Changes
- Inject CLI version during release.


.goreleaser.yaml
```diff
@@ -16,6 +16,7 @@ builds:
   - goos: linux
     goarch: arm64
   ldflags:
+  - -X github.com/bitrise-io/bitrise/version.VERSION={{ .Version }}
```
godrei, line 19: Shouldn't `github.com/bitrise-io/bitrise/version.VERSION=` be `github.com/bitrise-io/stepman/version.VERSION=`

_tests/integration/step_info_test.go
```diff
@@ -94,7 +94,7 @@ func TestStepInfoExitCode(t *testing.T) {
...
...slug\n","summary":"","category":"","is_required":true,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false}},{"opts":{"is_expand":true,"skip_if_empty":false,"title":"Xamarin.iOS License","description":"Set to yes if you want to download the Xamarin.iOS license file\n","summary":"","category":"","value_options":["yes","no"],"is_required":true,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false},"xamarin_ios_license":"yes"},{"opts":{"is_expand":true,"skip_if_empty":false,"title":"Xamarin.Android License","description":"Set to yes if you want to download the Xamarin.Android license file\n","summary":"","category":"","value_options":["yes","no"],"is_required":true,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false},"xamarin_android_license":"yes"},{"opts":{"is_expand":true,"skip_if_empty":false,"title":"Xamarin.Mac License","description":"Set to yes if you want to download the Xamarin.Mac license file\n","summary":"","category":"","value_options":["yes","no"],"is_required":true,"is_dont_change_value":false,"is_template":false,"is_sensitive":false,"unset":false},"xamarin_mac_license":"no"}]}`
```
godrei, line 97: Is the only change here the extra `"no_output_timeout":-1` ?

_tests/integration/version_test.go
```diff
@@ -14,7 +14,7 @@ func Test_VersionOutput(t *testing.T) {
 	{
 		out, err := command.RunCommandAndReturnCombinedStdoutAndStderr(binPath(), "version")
 		require.NoError(t, err, out)
-		require.Equal(t, "0.13.2", out)
+		require.Equal(t, "0.0.0-development", out)
```
godrei, line 17: shouldn't it be `0.13.2` ?

models/models.go
```diff
@@ -105,9 +105,11 @@ type StepModel struct {
 	//  steps will run which are marked with IsAlwaysRun.
 	IsSkippable *bool `json:"is_skippable,omitempty" yaml:"is_skippable,omitempty"`
 	// RunIf : only run the step if the template example evaluates to true
-	RunIf   *string                `json:"run_if,omitempty" yaml:"run_if,omitempty"`
-	Timeout *int                   `json:"timeout,omitempty" yaml:"timeout,omitempty"`
-	Meta    map[string]interface{} `json:"meta,omitempty" yaml:"meta,omitempty"`
+	RunIf   *string `json:"run_if,omitempty" yaml:"run_if,omitempty"`
+	Timeout *int    `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	// The timeout (in seconds) until a Step with no output (stdout/stderr) is aborted
+	NoOutputTimeout *int64                 `json:"no_output_timeout,omitempty" yaml:"no_output_timeout,omitempty"`
```
godrei, line 111: As far as I see, the default value here is `-1`, so this property will be included by default / with the default value.

models/models_methods_test.go
```diff
@@ -65,6 +65,10 @@ func TestValidate(t *testing.T) {
 	require.EqualError(t, step.Audit(), "Invalid step: missing or empty required 'source.commit' property")
 	step.Source.Commit = "1e1482141079fc12def64d88cb7825b8f1cb1dc3"
 
+	step.NoOutputTimeout = new(int64)
+	*step.NoOutputTimeout = -1
+	require.EqualError(t, step.Audit(), "Invalid step: no output timeout less then 0")
```
godrei, line 70: -1 is the default

models/models.go
godrei, line 111: why not simple int is the type?

_tests/integration/version_test.go
ofalvai: The final version number gets injected at build time based on the git tag of the release.

models/models_methods.go
```diff
@@ -176,6 +178,9 @@ func (step *StepModel) AuditBeforeShare() error {
 	if step.Timeout != nil && *step.Timeout < 0 {
 		return errors.New("Invalid step: timeout less then 0")
 	}
+	if step.NoOutputTimeout != nil && *step.NoOutputTimeout < 0 {
+		return errors.New("Invalid step: no output timeout less then 0")
```
ofalvai, line 182:
```suggestion
		return errors.New("Invalid step: no_output_timeout is less than 0")
```

_tests/integration/step_info_test.go
lpusok: Yes.

version/version.go
```diff
@@ -1,4 +1,4 @@
 package version
 
-// VERSION ...
-const VERSION = "0.13.2"
+// Version is the stepman version number. It's defined at build time using -ldflags
+var Version = "0.0.0-development"
```
lpusok, line 4: Removed const to allow value to be injected by linker.

.goreleaser.yaml
lpusok: Fixed.

models/models.go
lpusok: Changed default to 0. -1 is accepted still, meaning timeout is disabled.

models/models_methods.go
lpusok: Fixed.

models/models.go
```diff
@@ -105,9 +105,12 @@ type StepModel struct {
 	//  steps will run which are marked with IsAlwaysRun.
 	IsSkippable *bool `json:"is_skippable,omitempty" yaml:"is_skippable,omitempty"`
 	// RunIf : only run the step if the template example evaluates to true
-	RunIf   *string                `json:"run_if,omitempty" yaml:"run_if,omitempty"`
-	Timeout *int                   `json:"timeout,omitempty" yaml:"timeout,omitempty"`
-	Meta    map[string]interface{} `json:"meta,omitempty" yaml:"meta,omitempty"`
+	RunIf   *string `json:"run_if,omitempty" yaml:"run_if,omitempty"`
+	Timeout *int    `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	// The timeout (in seconds) until a Step with no output (stdout/stderr) is aborted
+	// Default: 0; Timeout disabled: -1
```
godrei, line 111: Timeout disabled: -1 -> do we need to handle this case separately?
AFAIK the old timeout is used if > 0. Should not we simply do the same with this timeout?

ofalvai: I agree, I think it means the same to have the timeout disabled or not having a timeout at all.

lpusok: Agreed to do not use 0 as the default, so we can differentiate between the value being nil (not set) and disabled (set to 0).

godrei (approved):

PR #323, opened 2024-05-16
[ACI-2048] Add offline mode

lpusok:
- Added activate support for offline mode
- Added preload
- Added listing for preloaded steps (using the unused short logging mode)

### DIFF
--- a/cli/activate.go
+++ b/cli/activate.go
@@ -189,7 +189,7 @@ func listCachedStepVersion(log stepman.Logger, stepLib models.StepCollectionMode
 		versions = append(versions, v)
 	}
 
-	slices.SortFunc(versions, models.LessSemver)
+	slices.SortFunc(versions, models.CmpSemver)
 
 	versionsStr := make([]string, len(versions))
 	for i, v := range versions {
--- a/cli/activate.go
+++ b/cli/activate.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 
 	"github.com/bitrise-io/go-utils/command"
 	"github.com/bitrise-io/go-utils/log"
@@ -70,12 +71,14 @@ func activate(c *cli.Context) error {
 	version := c.String(VersionKey)
 	copyYML := c.String(CopyYMLKey)
 	update := c.Bool(UpdateKey)
+	logger := log.NewDefaultLogger(false)
+	isOfflineMode := false
 
-	return Activate(stepLibURI, id, version, path, copyYML, update, log.NewDefaultLogger(false))
+	return Activate(stepLibURI, id, version, path, copyYML, update, logger, isOfflineMode)
 }
 
 // Activate ...
-func Activate(stepLibURI, id, version, destination, destinationStepYML string, updateLibrary bool, log stepman.Logger) error {
+func Activate(stepLibURI, id, version, destination, destinationStepYML string, updateLibrary bool, log stepman.Logger, isOfflineMode bool) error {
 	stepLib, err := stepman.ReadStepSpec(stepLibURI)
 	if err != nil {
 		return fmt.Errorf("failed to read %s steplib: %s", stepLibURI, err)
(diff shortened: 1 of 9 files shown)


preload/preload_steps.go
```diff
@@ -0,0 +1,240 @@
...
+
+	"github.com/bitrise-io/go-utils/colorstring"
+	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/bitrise-io/stepman/models"
+	"github.com/bitrise-io/stepman/stepman"
+)
+
+const (
+	workers       = 10
+	monthDuration = 24 * time.Hour * 31
```
ofalvai, line 17: I'm pretty sure it's not a good idea 😄 

I think we should take the current wall clock time, subtract `PatchesSinceMonths`, then compare that datetime to the step publish datetime.

preload/preload_steps.go
```diff
@@ -0,0 +1,240 @@
...
+	stepID  string
+	version string
+	status  string
+	err     error
+}
+
+// PreloadBitriseSteps preloads the cache with Bitrise owned steps
+func PreloadBitriseSteps(log stepman.Logger, steplibURL, maintaner string, opts PreloadOpts) error {
+	if opts.UseBinaryExecutable {
+		return fmt.Errorf("binary steps are not supported yet")
```
ofalvai, line 43: I'm afraid merging it like this, why don't we just remove `opts.UseBinaryExecutable` and add it in the next PR once it's working?

preload/preload_steps.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Preload latest version
+	latestVersionNumber := step.LatestVersionNumber
+	latestVersion, found := step.LatestVersion()
+	if !found {
+		return results, fmt.Errorf("failed to find latest version for step %s", stepID)
+	}
+
+	log.Infof("Preloading step %s", stepID)
```
ofalvai, line 161:
```suggestion
	log.Infof("Preloading step %s@latest", stepID)
```
WDYT?

models/version_constraint.go
```diff
@@ -42,6 +42,31 @@ func parseSemver(version string) (Semver, error) {
 	}, nil
 }
 
+func LessSemver(a, b Semver) int {
```
ofalvai, line 45: I'd call it `CmpSemver`, just to follow the Go stdlib conventions

cli/activate.go
```diff
@@ -70,12 +71,14 @@ func activate(c *cli.Context) error {
...
+	logger := log.NewDefaultLogger(false)
+	isOfflineMode := false
 
-	return Activate(stepLibURI, id, version, path, copyYML, update, log.NewDefaultLogger(false))
+	return Activate(stepLibURI, id, version, path, copyYML, update, logger, isOfflineMode)
 }
 
 // Activate ...
-func Activate(stepLibURI, id, version, destination, destinationStepYML string, updateLibrary bool, log stepman.Logger) error {
+func Activate(stepLibURI, id, version, destination, destinationStepYML string, updateLibrary bool, log stepman.Logger, isOfflineMode bool) error {
```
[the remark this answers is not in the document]
lpusok: Correct.

preload/preload_steps.go
lpusok: Ok, can do that.

preload/preload_steps.go
lpusok: It seems scary yes. Here is the comparison:
https://github.com/bitrise-io/stepman/blob/f66b89dccc9a8fc96684c81276fb818ba2365406/preload/preload_steps_util.go#L22
Went with this as wanted a simple way to parametrize and month seemed like a usable unit of time. I think could switch to days.
Could pass in a duration as well (then will have to do the conversion from days to duration at a level above).

lpusok: OK I now see that there is https://pkg.go.dev/time#Time.AddDate. Which in turn is timezone dependent :0.

ofalvai: No, I'm okay with this month param, I just don't want to hardcode 1 month to N seconds because what we usually want is to go back N calendar months in the local timezone and at the current date.

ofalvai (approved):

PR #328, opened 2024-06-28
Move git step activator from CLI to stepman

ofalvai:
See: https://github.com/bitrise-io/bitrise/issues/973

### DIFF
--- a/cli/activate.go
+++ b/cli/activate.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"slices"
 
+	"github.com/bitrise-io/bitrise/configs"
 	"github.com/bitrise-io/go-utils/command"
 	"github.com/bitrise-io/go-utils/log"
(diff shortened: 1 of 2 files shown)


cli/activate.go
```diff
@@ -98,7 +98,7 @@ func Activate(stepLibURI, id, version, destination, destinationStepYML string, u
 				versionList = versionList + fmt.Sprintf("\n- %s", version)
 			}
 
-			errMsg := fmt.Sprintf("version is not available in the local cache and $BITRISE_BETA_OFFLINE_MODE is set. %s", versionList)
+			errMsg := fmt.Sprintf("version is not available in the local cache and $BITRISE_OFFLINE_MODE is set. %s", versionList)
```
ofalvai, line 101: Spotted this earlier. We can't import the env key from the CLI codebase because that would be a circular dependency

lpusok (approved):

PR #332, opened 2024-07-05
Clean up activators

ofalvai:
Context: https://github.com/bitrise-io/bitrise/issues/973

Matching PR in the CLI repo: https://github.com/bitrise-io/bitrise/pull/981

### DIFF
--- a/activator/activator.go
+++ b/activator/activator.go
@@ -3,14 +3,6 @@ package activator
 type ActivatedStep struct {
 	StepYMLPath string
 
-	// TODO: this is a mess and only makes sense in the context of a path:: ref
-	// This should be cleaned up when all step actions are moved here from the CLI,
-	// but I don't want to blow up my PR with that change.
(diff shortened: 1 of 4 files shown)


activator/activator.go
```diff
@@ -3,14 +3,6 @@ package activator
 type ActivatedStep struct {
 	StepYMLPath string
 
-	// TODO: this is a mess and only makes sense in the context of a path:: ref
-	// This should be cleaned up when all step actions are moved here from the CLI,
-	// but I don't want to blow up my PR with that change.
-	OrigStepYMLPath string
-
-	// TODO: is this always the same as the `activatedStepDir` function param in all activations? Can we clean this up?
-	WorkDir string
```
ofalvai, line 12: It's not used in the CLI at the moment, so I removed it. This might be needed again when we add binary step support, but we'll see.

tothszabi (approved):

PR #333, opened 2024-07-24
Remove activate subcommand, move code out of `cli` package

ofalvai:
Ref: https://github.com/bitrise-io/bitrise/issues/973

Because step activation is no longer used as a stepman subcommand (the Bitrise CLI imports it as a Go lib), we can move the logic out of the huge `cli` package and delete some code.

The new organization isn't ideal either, but I ran into import cycles and had to compromise on this.


activator/steplib/activate.go
```diff
@@ -118,17 +54,9 @@ func Activate(stepLibURI, id, version, destination, destinationStepYML string, u
 	return nil
 }
 
-func queryStep(stepLib models.StepCollectionModel, stepLibURI string, id, version string, updateLibrary bool, log stepman.Logger) (models.StepModel, string, error) {
```
ofalvai, line 121: `updateLibrary` was only called with `false` (except for a test), so I cleaned it up

tothszabi (approved):

PR #334, opened 2024-08-27
Activate step executable

ofalvai:


activator/steplib/activate.go
```diff
@@ -4,57 +4,51 @@ import (
...
 	"slices"
+	"time"
 
 	"github.com/bitrise-io/go-utils/command"
 	"github.com/bitrise-io/go-utils/pathutil"
 	"github.com/bitrise-io/stepman/models"
 	"github.com/bitrise-io/stepman/stepman"
 )
 
-var errStepNotAvailableOfflineMode error = fmt.Errorf("step not available in offline mode")
```
ofalvai, line 15: Inlined this error handling into `activateStepSource()` because offline mode needs to be reworked for step executables anyway (we only cache the step source now, we'll need to cache the executables themselves)

activator/steplib/activate.go
```diff
@@ -75,46 +69,6 @@ func queryStep(stepLib models.StepCollectionModel, stepLibURI string, id, versio
 	return step, version, nil
 }
 
-func activateStep(stepLib models.StepCollectionModel, stepLibURI, id, version string, step models.StepModel, log stepman.Logger, isOfflineMode bool) (string, error) {
```
ofalvai, line 78: ➡️ `activate_source.go`

activator/steplib/activate.go
```diff
@@ -138,8 +92,14 @@ func copyStepYML(libraryURL, id, version, dest string) error {
 func ListCachedStepVersions(log stepman.Logger, stepLib models.StepCollectionModel, stepLibURI, stepID string) []string {
 	versions := []models.Semver{}
 
-	for version, step := range stepLib.Steps[stepID].Versions {
-		_, err := activateStep(stepLib, stepLibURI, stepID, version, step, log, true)
```
ofalvai, line 142: This was doing a full step activation just to check the error. I replaced this with a proper stepman call

models/models.go
```diff
@@ -120,6 +122,15 @@ type StepGroupModel struct {
 	Versions            map[string]StepModel `json:"versions,omitempty" yaml:"versions,omitempty"`
 }
 
+// Key: platform, as in runtime.GOOS + runtime.GOARCH
+// Examples: darwin-arm64, linux-amd64
+type Executables map[string]Executable
+
+type Executable struct {
+	Url string `json:"url,omitempty" yaml:"url,omitempty"`
```
lpusok, line 130: I am not convinced it is a good thing to add the URL in the step.ym, as we could keep it less transparent. For example we can host binaries in different storage providers, caches.

An other complication would be if we decide to upgrade go version and recompile the stepsm then the hash could change. Not sure if we want to do this tough.

activator/steplib/activate.go
ofalvai: Oh, true. The reason I had to replace this is the circular dependency between the `activateStep()` function that now calls this function.

models/models.go
ofalvai: Forgot to mention that this is only present in the steplib-submitted `step.yml`, just like the `source` block (which is not a great design, but it is what it is). So I imagine this `executable` block to be filled by the steplib submission tooling, just like it happens today with the `source` block. This way, a given URL is only associated with a tagged step release, not the step repo itself.

ofalvai: > An other complication would be if we decide to upgrade go version and recompile the stepman then the hash could change. Not sure if we want to do this tough.

I don't get this part though. Upgrade the go version where exactly?

ofalvai: I imagine the Go compiler upgrade like this:
- step is at v1.0.0 -> it has a tagged release URL and a matching hash
- Go upgrade, and any other tooling changes in the step repo
- Cut a new step release at v1.0.1 -> new tagged release URL and a new hash

viktorbenei:
[View build log](https://app.bitrise.io/build/f80525f4-e8b2-4354-91fb-d57590107af3)

viktorbenei:
[View build log](https://app.bitrise.io/build/dd182f09-c558-48d9-b633-4502a2223c71)

viktorbenei:
[View build log](https://app.bitrise.io/build/4369fe36-9107-425a-bc72-cd40d205f62c)

viktorbenei:
[View build log](https://app.bitrise.io/build/e5ae099e-8ef3-466c-a32e-47054401837b)

lpusok (dismissed):

viktorbenei:
[View build log](https://app.bitrise.io/build/b09da42a-5a92-4a00-829c-635cd8f6a8dd)

marcell-vida (approved):

PR #339, opened 2025-03-05
Update golang_min_version.go

lpusok:
Update to minimum go version 1.21.10.

### DIFF
--- a/toolkits/golang_min_version.go
+++ b/toolkits/golang_min_version.go
@@ -1,5 +1,5 @@
 package toolkits
 
 const (
-	minGoVersionForToolkit = "1.20.8"
+	minGoVersionForToolkit = "1.21.10"
 )


toolkits/golang_min_version.go
```diff
@@ -1,5 +1,5 @@
 package toolkits
 
 const (
-	minGoVersionForToolkit = "1.20.8"
+	minGoVersionForToolkit = "1.21.10"
```
godrei, line 4: We update to this version because `slices` was introduced in Go 1.21 (based on: [this article](https://www.gopherguides.com/articles/golang-slices-package)) right?

lpusok: Yes and this version is available on all not self hosted stacks as well.

godrei (approved):

PR #369, opened 2026-04-24
Fallback download URL logic for steps

marcell-vida:


activator/steplib/activate.go
```diff
@@ -15,8 +15,13 @@ import (
 )
 
 const precompiledStepsEnv = "BITRISE_EXPERIMENT_PRECOMPILED_STEPS"
-const precompiledStepsDefaultStorage = "https://storage.googleapis.com/bitrise-steplib-storage"
-const precompiledStepsPrimaryStorageEnv = "BITRISE_PRECOMPILED_STEPS_PRIMARY_STORAGE"
+const precompiledStepsStorageURLsEnv = "BITRISE_PRECOMPILED_STEPS_STORAGE_URLS"
```
ofalvai, line 18: I don't think we'll ever change the URL of the backup location (GCS bucket). I think this diff could be simpler if the backup location was hardcoded into this repo, and only the "override" would be coming from the env var.

Badlazzor (dismissed):

activator/steplib/activate.go
marcell-vida: I was mainly thinking about testing the new bucket, where we don't want any fallback options to be used. With this change we could just pass a list with 1 source and it could fail if something is misconfigured.

ofalvai: Thank you, that's a good point and I agree

activator/steplib/activate.go
```diff
@@ -15,8 +15,13 @@ import (
...
+const precompiledStepsPrimaryStorageEnvDeprecated = "BITRISE_PRECOMPILED_STEPS_PRIMARY_STORAGE"
```
ofalvai: It's not used anywhere, we could just kill it. The only external user of these env vars is this: https://github.com/bitrise-io/bitrise-website/blob/54af00872b6a6b690f9ed68945270d0f2436ac2c/components/ci/app/helpers/workflow_helper.rb#L342

activator/steplib/activate_executable.go
```diff
@@ -106,11 +100,61 @@ func validateHash(filePath string, expectedHash string) error {
...
+func downloadFromURLs(urls []string) (io.ReadCloser, error) {
+	var errs []error
+	for _, url := range urls {
+		resp, err := retryablehttp.Get(url)
+		if err == nil && resp.StatusCode < 400 {
+			return resp.Body, nil
+		}
+
+		if err != nil {
+			log.Warnf("Failed to download from %s: %s\n", url, err)
```
ofalvai:
```suggestion
			log.Warnf("Failed to download step from %s: %s\n", url, err)
```

activator/steplib/activate_executable_test.go
```diff
@@ -60,80 +63,166 @@ func TestValidateHash(t *testing.T) {
...
 			},
-			executable: models.Executable{
-				StorageURI: "steps/step1.tar.gz",
-			},
-			expectedURL: "https://custom.example.com/storage/steps/step1.tar.gz",
 		},
 		{
-			name: "With default base URL",
-			env:  map[string]string{},
+			name:        "Override list via env var",
```
ofalvai: nit: there is no env var involved in this test at all, we are testing a different layer

ofalvai (changes requested):
Just one more simplification idea, then it's ready to merge!

ofalvai (approved):

PR #376, opened 2026-06-02
feat: V2 inventory wire types + generator (steplibindex, indexgen)

lpusok:
## Summary

PR 1 of the V2 steplib series: the wire-format types and the generator that builds the V2 inventory tree from a bitrise-steplib clone — plus the shared test fixtures and CI/lint setup the rest of the series builds on.

## Wire types — `steplibrary/steplibindex`

- Single source of truth for the V2 JSON schema, shared by the generator (write path) and the steplibrary reader (read path) — no duplicated structs. Types: `Meta`, `StepInfo`, `Deprecation`, `StepIDs`, `LatestPointer`, `Versions`.
- **Stable wire shape:** index/per-step collections always render as `[]`/`{}` (never `null`); nullable values (`deprecation`, `published_at`) as `null`; optional inventory metadata (`steplib_commit_sha`, `steplib_source`, `download_locations`) uses `omitempty`.
- `FormatVersion` (= 2) and `VersionDir()` (→ `v2`) — the whole tree is rooted under the format-version dir so multiple versions can be hosted side by side.

## Generator — `steplibrary/indexgen`

- Reads a bitrise-steplib clone via `fs.FS` (`collect.go`), builds the derived index files (`index.go`), and emits everything through a byte/file-accounting `writer.go`. The tree is staged in a sibling temp dir and **published with a single atomic rename**, so a failed run never leaves a half-written inventory.
- Output layout, rooted under the format-version dir:
  ```
  v2/
  ├─ meta.json
  ├─ index/                         (derived indexes; regeneratable, short-TTL)
  │  ├─ step_ids.json
  │  └─ steps/<id>/{latest.json, versions.json}
  └─ steps/                         (source of truth; immutable per version)
     └─ <id>/{step-info.json, assets/…, <version>/step.json}
  ```
- `step-info.yml` is **required** per step (missing → error). Copied assets keep their source permissions; files the generator authors are owner-only (`0600`/`0700`).
- **Deterministic / reproducible:** option defaulting is centralized in `withDefaults` — `GeneratedAt` is normalized to UTC whole-second RFC3339, `SteplibCommitSHA` defaults to the clone's `HEAD`. Two runs over the same input produce byte-identical output.
- Exposed as `indexgen.Generate` (URI entry point); the lower-level `fs.FS` core stays unexported. The production CLI is the `bitrise steps generate-steplib` subcommand (STEP-2428, merges right after this PR) — no standalone tool ships here.

## Notable schema decisions

- `index/steps/<id>/versions.json` = `{ step_id, versions: [...] }` — a newest-first list of **version strings**. Per-version detail (commit, published_at) lives in `step.json`; the latest pointer in `latest.json`.
- `asset_urls` is a `[]string` of step-relative URLs.
- **Dropped** the fat `latest_versions.json` catalog and `has_executable` — both redundant (recorded in `STEP-2374-plan.md`).

## Shared fixtures & CI

- `internal/specfixtures` — a small real-world sample steplib (bash → Script, deprecated → branch-cache, multi-platform → deploy-to-bitrise-io, hello-step) embedded via `//go:embed` and exposed as an `fs.FS`, so every package's tests use the same fixtures regardless of working directory.
- CI (`bitrise run test`): YAML lint via steps-check's `lint` workflow (yamllint), `golangci-lint` with the repo's own `.golangci.yml` (incl. `exhaustruct`), and `go test ./...`.

## PRs in this series

- **PR 1 (this)** — wire types + inventory generator
- PR 1.5 — inventory validator + wire-format / determinism tests
- PR 2 — httpfetch client + steplibrary read path
- PR 3 — precompiled binary download + V1 source fallback
- PR 4 — activator wiring + CI

## Test plan

- [x] `go test ./steplibrary/steplibindex/... ./steplibrary/indexgen/...`
- [x] `bitrise run test` (YAML lint + golangci-lint + Go test)
- [ ] local end-to-end generation against a real steplib (manual; skipped in CI):
  ```
  RUN_STEPLIB_GEN=1 go test -run TestGenerate_local -v ./steplibrary/indexgen/
  ```
- [ ] e2e via `bitrise_specv2.yml`: `bitrise steps generate-steplib` → local HTTP host → real bitrise CLI with `BITRISE_EXPERIMENT_STEPLIB_V2=true` (requires STEP-2428)


.yamllint
ofalvai, line 1: Why is this file included in this PR?
Also, at this point we should not introduce new usages of yamllint, and import the pre-confogured yamlfmt workflow from the steps-check repo instead

steplibrary/steplibindex/steplibindex.go
ofalvai, line 1: What's the reason for introducing these types, instead of reusing the existing ones in this repo? I thought we agreed to reuse the existing YML file contents, and then later iterate on the fields if needed

steplibrary/specgen/internal/cmd/steplib-gen/main.go
ofalvai, line 1: Wouldn't E2E tests be able to import a high-level Go function and run everything through that? We could eliminate this CLI then

steplibrary/specgen/testdata/input/steps/bash-step/1.0.0/step.yml
ofalvai, line 1: Nit: a real-world step.yml could be a better test data, and it's just a copy-paste from the step library repo

ofalvai (changes requested):

.yamllint
lpusok: Purpose is to ignore contents of `steplibrary/specgen/testdata/`. This way it overrides the default config we used everywhere.
Added a [note]( https://bitrise.atlassian.net/browse/STEP-2431) to try to use new tooling as followup.
One problem may be that this is a different github org and I am not sure new config works.

steplibrary/specgen/internal/cmd/steplib-gen/main.go
lpusok: It is used from bitrise_specv2.yml, not go code.  We may rethink if that is the best way but I have no concrete ideas right now (other than maybe make this internal if possible, and remove all uneeded extra params)

steplibrary/steplibindex/steplibindex.go
lpusok: We reused step.yml. Most of these are new needed for version catalog. Everything is as agreed on.
The only ones that overlap are:
- Meta, but we added new fields (SteplibCommitSHA, UpdatedAt). FormatVersion is a single number now. DownloadLocations stayed, but the plan is to iterate on that part in the future so a new type is needed.
- StepInfo + Deprecation: which I think now has AssetURLs added and slightly different format. -> This is the single one that is converted back to the previous struct to be passed back to the CLI. Taking a [note](https://bitrise.atlassian.net/browse/STEP-2431) to clean it up inside the CLI codebase.
---
Not a huge mental overhead to have these as they are simple structs overall.

.yamllint
ofalvai: Why not format the testdata once, just to satisfy the current linter? Do these tests rely on malformed YML files?

steplibrary/steplibindex/steplibindex.go
ofalvai: I'm fine with the new types for the *projections*, just want to avoid duplicating what's already defined in the `models` package. So in practice:

1. `Meta`: got it, those new fields make sense! Do we really need `DownloadLocations` though? Is it going to be useful in this first iteration? If not, I'd just remove it for now, and only introduce it when we have a concrete plan for using it.
1. `StepInfo` + `Deprecation`: this is the one that I'd like to reuse from the models package, and if we are not satisfied with its current shape, then tweak the fields (instead of duplicating the struct)

steplibrary/specgen/internal/cmd/steplib-gen/main.go
ofalvai: Got it! I don't understand that workflow yet, will come back to this once I understand what it's doing and why it's needed

steplibrary/spec/schemas.md
```diff
@@ -0,0 +1,323 @@
+# V2 inventory — schemas reference
```
ofalvai: Do we really want to check this design doc into git? I'm not absolutely against it, but it has to have a good reason to exist, because it could become stale over time very easily

steplibrary/spec/spec.go
```diff
@@ -0,0 +1,108 @@
...
+type Catalog struct {
+	GeneratedAt      time.Time              `json:"generated_at"`
+	SteplibCommitSHA string                 `json:"steplib_commit_sha,omitempty"`
+	Steps            map[string]CatalogEntry `json:"steps"`
+}
+
+// CatalogEntry is one step's entry in Catalog.Steps. Asset URLs are
+// inventory-root-relative so consumers can resolve them against the base URL
+// the catalog was fetched from, without any hosting URL baked in.
+type CatalogEntry struct {
```
ofalvai: I ran the generator to see how big `latest_versions.json` is in the real world, and it's 381 KB...which is not that great, especially if we account for the growth of steplib itself.

Could you give a justification for why each field need to be included in this structure? `latest_version` is obvious at this point, but I'd like to be really critical about anything else. I don't think it's a good idea to fetch a half-megabyte JSON by every client. We can always add fields later, but won't be able to remove, so we should be careful now.

Also, even if one use-case or client needs a certain field, it doesn't mean it should be included in the `latest_versions.json` catalog file. Maybe the better solution is to query N files for N steps. I'd like to see actual numbers and use-cases before adding a field to this projection.

steplibrary/spec/spec.go
```diff
@@ -0,0 +1,108 @@
...
+	Latest   string         `json:"latest"`
+	Versions []VersionEntry `json:"versions"`
+}
+
+// VersionEntry is a single entry in Versions.Versions.
+type VersionEntry struct {
+	Version       string     `json:"version"`
+	PublishedAt   *time.Time `json:"published_at,omitempty"`
+	HasExecutable bool       `json:"has_executable"`
+	Commit        string     `json:"commit,omitempty"`
```
ofalvai: It's not immediately obvious if this is the SHA of the steplib or the step repo itself.

steplibrary/specgen/internal/cmd/steplib-gen/main.go
```diff
@@ -0,0 +1,79 @@
+// Command steplib-gen sets up a step library by its source URI (cloning it into
+// stepman's local cache if needed) and writes the V2 inventory tree to an output
+// directory.
+//
+//	steplib-gen -steplib <steplib-source-uri> -output <out-dir> [-commit-sha <sha>]
+//
+// See STEP-2374-plan.md and steplibrary/spec/schemas.md for the format being generated.
```
ofalvai: This is no longer accurate

steplibrary/specgen/generator.go
```diff
@@ -0,0 +1,173 @@
+// Package specgen generates the V2 step library inventory tree from a
+// bitrise-steplib source. The wire-format types it emits live in
+// steplibrary/spec.
+//
+// The work splits across files by phase:
+//   - generator.go — public API (Generate, GenerateFromSteplibClone) + orchestration
+//   - collect.go   — read the steplib source into []parsedStep
+//   - index.go     — build the derived spec/ index files and write step files
+//   - writer.go    — the file-emitting writer with byte/file accounting
```
ofalvai: nit: do we need such detailed comments? Any human/agent can just follow the references from this file to learn about the inner workings. On the other hand, these comments can easily go stale as those files get refactored.

steplibrary/specgen/generator.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return repo.RevParse("HEAD").RunAndReturnTrimmedCombinedOutput()
+}
+
+// GenerateFromSteplibClone reads a bitrise-steplib clone from inputFS and writes
+// the V2 inventory tree to outputDir. The tree is staged in a sibling temp
+// directory and published with a single rename on success, so a failure
+// mid-generation never leaves a half-written inventory at outputDir; any
+// existing tree at outputDir is replaced wholesale.
+func GenerateFromSteplibClone(inputFS fs.FS, outputDir string, opts Options, log stepman.Logger) (_ Stats, err error) {
```
ofalvai: If this is only exists for tests, can we make it private?

ofalvai: And possible move the whole function to a test file.

Edit: nevermind, I see how `Generate()` also calls it. But it could still be private

steplibrary/specgen/generator.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Default the recorded commit SHA to the checked-out library's HEAD when
+	// the caller didn't pin one.
+	if opts.SteplibCommitSHA == "" {
+		sha, err := headCommitSHA(libDir)
+		if err != nil {
+			return Stats{}, fmt.Errorf("resolve steplib HEAD commit: %w", err)
+		}
+		opts.SteplibCommitSHA = sha
+	}
```
ofalvai: Thoughts on moving this logic to `withDefaults()`? After all, we are handing the default value here.

steplibrary/spec/spec.go
```diff
@@ -0,0 +1,108 @@
...
+	HasExecutable   bool              `json:"has_executable"`
+	Deprecation     *Deprecation      `json:"deprecation"`
+}
+
+// LatestPointer is spec/steps/<id>/latest.json: per-step latest pointers.
+// Answers Latest and MajorLocked constraints in a single small fetch.
+type LatestPointer struct {
+	StepID        string            `json:"step_id"`
+	Latest        string            `json:"latest"`
+	LatestByMajor map[string]string `json:"latest_by_major,omitempty"`
```
ofalvai: About `omitempty`: is it even possible to have an empty value in this field?

steplibrary/steplibindex/steplibindex.go
lpusok: ```Do we really need DownloadLocations though? Is it going to be useful in this first iteration? If not, I'd just remove it for now, and only introduce it when we have a concrete plan for using it.```
it is used to download step.zip (src zip), as in the legacy code. So needed as we kinda need that functionality

steplibrary/specgen/internal/cmd/steplib-gen/main.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/bitrise-io/stepman/steplibrary/specgen"
+)
+
+type stderrLogger struct{}
```
ofalvai: Why not use `log.NewDefaultLogger()` like every existing top-level function in this repo?

steplibrary/steplibindex/steplibindex.go
lpusok: `tweak the fields ` I think that is a bad idea as we are now on version 2, and do not want to touch version 1 (the git steplib format)

.yamllint
lpusok: Yes the step.ymls are malformed (even the official ones)

steplibrary/steplibindex/steplibindex.go
ofalvai: Yeah, I never considered that, right :( So even if we updated the structs and release a CLI version that handles both serialization and deserialization of the new fields, there is still a serialized copy of old data in the steplib repo, right?

steplibrary/spec/spec.go
```diff
@@ -0,0 +1,108 @@
...
+// (steplibrary/specgen) and the read path (steplibrary).
+//
+// The V2 layout splits the inventory into two URL prefixes:
+//   - steps/  — source of truth, self-contained per step, immutable per version
+//   - spec/   — derived index files, regeneratable from steps/, short-TTL
+//
+// All files are JSON. Per-version step manifests (steps/<id>/<v>/step.json)
+// use models.StepModel directly; the types below describe the new
+// inventory-level and index files that have no V1 equivalent.
+package spec
```
ofalvai: nit: what do you think about calling it `index` instead? To me, spec is kind of meaningless, and the main point of these files under the `spec` subdir is that they are pre-computed indexes (unlike the `steps` subdir, which contains the original `step.yml` data for each step version)

steplibrary/indexgen/collect.go
```diff
@@ -0,0 +1,197 @@
...
+	version string
+	semver  models.Semver
+	model   models.StepModel
+}
+
+// latest returns the highest-semver version. Only valid for steps with at least
+// one version (collectSteps drops versionless steps before they reach here).
+func (s parsedStep) latest() parsedVersion { return s.versions[len(s.versions)-1] }
+
+func readSteplibYML(inputFS fs.FS) (models.StepCollectionModel, error) {
```
ofalvai, line 61: Isn't it a duplicate of https://github.com/bitrise-io/stepman/blob/34234fe61c7b32b747368c92874e7fc8dcdebc50/stepman/util.go#L95-L106?

.yamllint
ofalvai: But do they need to be malformed for out tests? Why not just format them once, and commit those files to testdata? :)

steplibrary/spec/spec.go
lpusok: This is covered in the plan, this is a reference of the latest step for front-end reasons.
It is not used by stepman itself, so it is 0 overhead.
Also, do not want to redesign the API at this point as it was accepted as-is, at least until we have some real world feedback.

steplibrary/specgen/collect.go
```diff
@@ -0,0 +1,197 @@
...
+	s := parsedStep{
+		id:          id,
+		info:        spec.StepInfo{Maintainer: "", Deprecation: nil, AssetURLs: nil},
+		hasInfoFile: false,
+		assetFiles:  nil,
+		versions:    nil,
+	}
+	stepDir := "steps/" + id
+
+	info, hasInfo, err := readStepGroupInfo(inputFS, stepDir+"/step-info.yml")
```
ofalvai: Is it even realistic to not have a `step-info.yml` file for each step? I know we are walking the filesystem of a git repo, so everything is possible, but if this is an invalid state that shouldn't happen, then why not raise a hard error instead? I think it would catch errors sooner and simplify this code a lot.

steplibrary/spec/spec.go
ofalvai: I disagree, this is the time to design the right API, and we can always add more fields later, but won't be able to remove them without making a breaking change. I'm not challenging the index file itself (`latest_versions.json`), just the fields that are present in the JSON. If there is real feedback about a missing field, we can easily add one. But we won't be able to fix half-megabyte JSON fetches later.

> It is not used by stepman itself, so it is 0 overhead.

To be clear, I was referring to the overhead of downloading this file, especially in the browser context, where it makes the page load slower.

steplibrary/specgen/collect.go
```diff
@@ -0,0 +1,197 @@
...
+	s.hasInfoFile = hasInfo
+
+	assetFiles, err := listAssets(inputFS, stepDir+"/assets")
+	if err != nil {
+		return s, fmt.Errorf("list assets for %s: %w", id, err)
+	}
+	s.assetFiles = assetFiles
+	if len(assetFiles) > 0 {
+		if s.info.AssetURLs == nil {
+			s.info.AssetURLs = make(map[string]string, len(assetFiles))
```
ofalvai: nit: why not initialize this `AssetURLs` map in the "constructor", at the top of the function?

.yamllint
lpusok: I can but also I used some "real-world" ymls as examples, which then no longer be real-world. But possible.
From a semantic point of view it is not really needed to "clobber" test data on every run (ymlfmt rewrites stuff on every run), which seems to get agents surprised as well.

steplibrary/specgen/collect.go
```diff
@@ -0,0 +1,197 @@
...
+		s.versions = append(s.versions, parsedVersion{version: sub.Name(), semver: sv, model: model})
+	}
+
+	sort.Slice(s.versions, func(i, j int) bool {
+		return models.CmpSemver(s.versions[i].semver, s.versions[j].semver) < 0
+	})
+	return s, nil
+}
+
+func collectSteps(inputFS fs.FS, log stepman.Logger) ([]parsedStep, error) {
```
ofalvai: super minor nit: I know this file only has private functions, but I still like to put the "higher-level" functions on top (which are called from other files in the same package). I'm not sure LLMs write code this way.

steplibrary/indexgen/indexgen_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+// runGenerateFromSteplibClone runs the generator against the checked-in
+// testdata using a fresh temp output dir. Returns the output dir for assertions.
+func runGenerateFromSteplibClone(t *testing.T) string {
+	t.Helper()
+	out := t.TempDir()
+	_, gotErr := GenerateFromSteplibClone(
+		specfixtures.SteplibClone(),
+		out,
+		Options{GeneratedAt: fixedTime, SteplibCommitSHA: "deadbeefcafef00d"},
```
ofalvai, line 35: > deadbeefcafef00d

Noice!

.yamllint
ofalvai: Makes sense, but I think our tests are not there to verify whitespace and other syntactic differences. As long as they are valid YML and can be parsed, they are good test data.

steplibrary/steplibindex/steplibindex.go
lpusok: basically yes. We may need a migration plan (what to keep or not). But older CLI versions still depend on those fields. Adding fields is possible (if format_version is not validated) to the step info, that would not cause problems if we keep track of it

steplibrary/specgen/generator.go
```diff
@@ -0,0 +1,173 @@
+// Package specgen generates the V2 step library inventory tree from a
+// bitrise-steplib source. The wire-format types it emits live in
+// steplibrary/spec.
+//
+// The work splits across files by phase:
+//   - generator.go — public API (Generate, GenerateFromSteplibClone) + orchestration
+//   - collect.go   — read the steplib source into []parsedStep
+//   - index.go     — build the derived spec/ index files and write step files
+//   - writer.go    — the file-emitting writer with byte/file accounting
+package specgen
```
ofalvai: nit: I see this `generator.go` is the main file of this package, which calls functions from other files like `index.go` and `writer.go`. I'd rename this file to `specgen.go`, I think that's the Go convention when a module has multiple files.

steplibrary/spec/spec.go
lpusok: Agreed to remove endpoint completely.

steplibrary/spec/spec.go
lpusok: Removing HasExecutable and Commit fields for now as they are unused.

steplibrary/indexgen/collect.go
lpusok: it is, but this has fs.Fs instead of path so no longer compatible.

steplibrary/specgen/collect.go
lpusok: The default order was kind of the opposite, this was an explicit ask from me as a refactoring effort (based on some vauge recollection of a guideline). Probably would not have it done by hand tough

steplibrary/indexgen/collect.go
ofalvai: The conversion is only a `dirPath := os.DirFS(path)`, so I'd rather refactor `ParseStepCollection()` to take an `fs.FS` as the argument

steplibrary/specgen/collect.go
lpusok: We kinda flagged this earlier and decided the same, but maybe not fully applied

.yamllint
lpusok: linted tesdata

steplibrary/specgen/internal/cmd/steplib-gen/main.go
lpusok: Removed it, will be added in bitrise repo.

steplibrary/spec/schemas.md
lpusok: no

steplibrary/spec/spec.go
lpusok: renamed
 - spec → steplibindex (steplibrary/spec/ → steplibrary/steplibindex/)
  - specgen → indexgen (steplibrary/specgen/ → steplibrary/indexgen/)

steplibrary/specgen/collect.go
lpusok: Added validation.

steplibrary/specgen/collect.go
lpusok: converted to []string (map was not needed)

steplibrary/specgen/generator.go
lpusok: done

steplibrary/specgen/generator.go
lpusok: tried to simplify some

steplibrary/spec/spec.go
lpusok: no, as steps with 0 version is ignored with a warning. removed.

steplibrary/specgen/generator.go
lpusok: done.

steplibrary/indexgen/collect.go
lpusok: left as is, we can come back to this. Also wanted to minimize refactors of existing code.

steplibrary/specgen/generator.go
lpusok: renamed entry files to match package

steplibrary/specgen/internal/cmd/steplib-gen/main.go
lpusok: removed steplib-gen

steplibrary/indexgen/indexgen.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// generateFromSteplibClone reads a bitrise-steplib clone from inputFS and writes
+// the V2 inventory tree to outputDir. The tree is staged in a sibling temp
+// directory and published with a single rename on success, so a failure
+// mid-generation never leaves a half-written inventory at outputDir; any
+// existing tree at outputDir is replaced wholesale.
+func generateFromSteplibClone(inputFS fs.FS, outputDir string, opts Options, log stepman.Logger) (_ Stats, err error) {
+	start := time.Now()
+	// No git dir here (fs.FS source); SteplibCommitSHA is defaulted by Generate.
```
ofalvai, line 100: nit: sounds like a comment that only makes sense in the context of our recent refactors, I'd simply remove it

steplibrary/indexgen/collect.go
```diff
@@ -0,0 +1,196 @@
...
+
+// parsedVersion is a single step version with its semver parsed once at collect
+// time, so the write/index phase never re-parses the version string.
+type parsedVersion struct {
+	version string
+	semver  models.Semver
+	model   models.StepModel
+}
+
+// latest returns the highest-semver version. Only valid for steps with at least
```
ofalvai, line 32: nit: this is kinda stupid and pointless, a step cannot exist with zero versions, no?

ofalvai (approved):

PR #377, opened 2026-06-02
feat(steplibrary): V2 read path — httpfetch client + steplibrary reader

lpusok:
## Summary

The V2 read path: an HTTP client (`internal/httpfetch`) plus the `steplibrary` reader that fetches the sharded V2 inventory over HTTP and resolves + activates a step from it.

> **Note:** `Steplib.Activate` here is an intentional intermediate state — it resolves a version and writes `step.yml`, but does **not** yet download step code. PR 3 extends it with precompiled-binary + V1 source fallback.

Builds on the now-merged generator (#376) and validator (#381). Every fetch URL is built from `steplibindex` — the single source of truth for the inventory layout, shared with the generator and validator. Path construction is unified behind a `steplibindex.Path` type that carries both forms and exposes `.FS()` (the generator's `io/fs` writes) and `.URL()` (the reader's HTTP fetches, with per-segment escaping), so the FS and URL representations can't drift.

**Read path + activation (`steplibrary`, `internal/httpfetch`)**
- `HTTPAPI` fetches the V2 inventory over HTTP (`v2/index/…`, `v2/steps/…`); `httpfetch.Client` does atomic temp-file + rename downloads, with `DownloadWithHash` verifying sha256 *before* the rename so a partial or mismatched file never lands at the destination.
- `Steplib.Activate`: version resolution (latest / fixed / major-locked / minor-locked) + step-model fetch + `step.yml` write. Minor-lock selection reuses a shared `models.HighestForMajorMinor` kernel (the same "highest patch within major.minor" scan the V1 resolver does). The v2 `asset_urls` (`[]string`) is flattened back into the V1 `{filename → url}` map so the rest of the codebase keeps the same model shape.

**Activation result type (`steplibrary`)**
- Moves `ActivatedStep` / `ActivationType` out of `activator` into the `steplibrary` package (their canonical owner); callers reference `steplibrary.ActivatedStep` directly — no alias shim.
- `ActivatedStep` carries a `StepInfo` field, now populated by every activation path rather than via a caller-supplied pointer: `Steplib.Activate` (steplib, including the deprecation `GroupInfo`), and git/path-ref activation, which fill it from the local `step.yml`. Git/path reuse stepman's own builders — `stepman.QueryStepInfoFromPath` and the extracted `QueryStepInfoFromGitStepDir` (so the already-cloned dir isn't re-cloned) — matching the field conventions the bitrise CLI's `newStepInfoPtr` already uses.

**Tests**
- Unit coverage across resolver, routing, and the HTTP API (mocks point at the `v2/index` layout).
- Hermetic activation-result tests: path-ref activation from a temp dir, and git-ref activation from a locally-initialized repo (no network), asserting the populated `StepInfo`.
- A true **generator → HTTP → reader** integration test generates a V2 tree from the embedded `specfixtures` steplib, serves it over `httptest`, and exercises `HTTPAPI` — no checked-in inventory fixture to drift. It drives the unexported `fs.FS` generator via a test-only hook (`GenerateFromSteplibCloneForTest`), so the production API isn't widened.

### Not yet implemented / follow-ups
- V1-compatible source fetch via `stepman.DownloadStep` (zip + git fallback, retries, commit-hash verification) into the existing V1 on-disk cache. → PR 3
- **precompiled-binary** activation with multi-URL fallback + hash verification, transparent fallback to **source**. → PR 3
- **Offline mode** — `isOfflineMode` is threaded through, but the V2 inventory reads (`HTTPAPI`) always hit the network. → later
- **Local caching** — no on-disk cache / `ETag` revalidation for the V2 inventory JSON yet; every read is a fresh HTTP fetch. → PR 5
- Retiring the `stepInfoPtr` workaround in `activator/steplib_ref` now that `ActivatedStep.StepInfo` is populated — needs a coordinated bitrise CLI change. → later

## PRs in this series
- PR 1 (#376, merged) — V2 wire types + inventory generator
- PR 1.5 (#381, merged) — inventory validator + wire-format / determinism tests
- **PR 2 (this)** — httpfetch client + steplibrary read path
- PR 3 (#378) — precompiled binary download + V1 source fallback
- PR 4 (#379) — activator wiring + CI
- PR 5 (#380) — HTTP cache

## Test plan
- [x] `go test ./internal/httpfetch/... ./steplibrary/... ./activator/... ./models/...`
- [x] generator → HTTP → reader integration test runs against a local test server


activator/result/result.go
```diff
@@ -0,0 +1,33 @@
...
+	// - step was activated from a git reference (we checked out the source dir directly)
+	// - step was activated from a local path (we copied the source dir directly)
+	// - step was activated from a steplib reference, but step.yml has no entry for pre-compiled binaries (we fallback to source checkout)
+	// - step was activated from a stpelib reference, but step.yml has no pre-compiled binary for the current OS+arch combo (we fallback to source checkout)
+	ExecutablePath string
+
+	// DidStepLibUpdate indicates that the local steplib cache was updated while resolving the exact step version.
+	// TODO: this is a leaky abstraction and we shouldn't signal this here, but it requires a bigger refactor.
+	// (stepman should keep track of this info in a file probably)
+	DidStepLibUpdate bool
```
godrei, line 23: Do we really need it here?

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+	// the final path.
+	Download(ctx context.Context, destPath, url string) error
+	// DownloadWithHash behaves like Download but also verifies that the
+	// downloaded content matches expectedHash ("sha256-<hex>"). The temp file
+	// is removed and an error is returned if the hash does not match, so a
+	// mismatched file never appears at destPath.
+	DownloadWithHash(ctx context.Context, destPath, url, expectedHash string) error
+}
+
+// Logger is the minimal logging interface required by Client: the retry
```
godrei: I'm not sure if these lengthy comments are needed. For example in this case:
- a few lines below it is already clear who uses this struct
- it is pretty obvious that any struct with `Debugf(format string, v ...any)` satisfies the interface

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("build request for %s: %w", url, err)
+	}
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("GET %s: %w", url, err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, errors.Join(
+			fmt.Errorf("GET %s: unexpected status %d", url, resp.StatusCode),
```
godrei: Don't we need the response body in case of error?

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := os.MkdirAll(filepath.Dir(destPath), 0o755); err != nil {
+		return fmt.Errorf("create dest dir for %s: %w", destPath, err)
+	}
+
+	tmpPath, _, err := c.fetchToTemp(ctx, filepath.Dir(destPath), url)
+	if err != nil {
+		return err
+	}
+
+	if renameErr := os.Rename(tmpPath, destPath); renameErr != nil {
```
godrei: Why don't we simply use `err` instead of `renameErr`?

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+		return errors.Join(fmt.Errorf("rename %s to %s: %w", tmpPath, destPath, renameErr), os.Remove(tmpPath))
```
godrei: I guess `os.Remove(tmpPath)` is a cleanup, it took some time to understand the returned error here. Can't we simply cleanup the tmp dir in a defer statement right after creating it?

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return nil
+}
+
+func (c *client) DownloadWithHash(ctx context.Context, destPath, url, expectedHash string) error {
```
godrei, line 109: Download and DownloadWithHash only differs in has validation yet many code is duplicated, can't we share a bigger chunk of code?

steplibrary/steplibrary.go
```diff
@@ -0,0 +1,41 @@
...
+	if err == nil {
+		stepYML, err = yaml.Marshal(stepModel)
+		if err != nil {
+			err = fmt.Errorf("marshal step model to YAML: %w", err)
+		}
+	}
+	if err == nil {
+		err = s.fileManager.WriteBytes(outputPaths.YMLPath, stepYML)
+	}
+	if err != nil {
```
godrei, line 48: Why don't we have this error handling right after `s.getStepVersionInfo` ?

steplibrary/steplibrary.go
```diff
@@ -0,0 +1,41 @@
...
+		return result.ActivatedStep{}, err
+	}
+
+	return result.ActivatedStep{
+		StepInfo:         stepInfo,
+		StepYMLPath:      outputPaths.YMLPath,
+		ExecutablePath:   "",
+		ActivationType:   result.ActivationTypeSteplibSource,
+		DidStepLibUpdate: false, // deprecated
```
godrei, line 61: Can't we just remove this field?

steplibrary/http_api.go
```diff
@@ -0,0 +1,90 @@
...
+func NewHTTPAPI(baseURL string, fetcher httpfetch.Client) *HTTPAPI {
+	return &HTTPAPI{
+		BaseURL: strings.TrimRight(baseURL, "/"),
+		Fetcher: fetcher,
+	}
+}
+
+func (h *HTTPAPI) GetAllStepIDs(ctx context.Context) ([]string, error) {
+	var payload spec.StepIDs
+	if err := h.fetchJSON(ctx, "/spec/step_ids.json", &payload); err != nil {
```
godrei: Haven't we introduce exported functions for these paths in a previous PR?

steplibrary/resolve.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+func (s *Steplib) getStepVersionInfo(ctx context.Context, stepID, version string) (models.StepInfoModel, ResolvedStepVersion, error) {
+	var err error
+	if stepID == "" {
+		err = errors.New("missing required input: step id")
+	}
+
+	var allSteps []string
+	if err == nil {
```
godrei: Why is this strange error handling?

steplibrary/resolve.go
```diff
@@ -0,0 +1,153 @@
...
+
+// resolveMinorLocked picks the highest patch within `versions` matching the
+// constraint's Major+Minor. Unparseable entries are skipped.
+func resolveMinorLocked(versions []string, constraint models.Semver) (string, error) {
+	var best models.Semver
+	found := false
+	for _, raw := range versions {
+		sv, err := models.ParseSemver(raw)
+		if err != nil {
+			continue
```
godrei: Why are they skipped? Why not an error?

steplibrary/steplib.go
```diff
@@ -0,0 +1,34 @@
+package steplibrary
```
godrei, line 1: I guess this file just a preparation for the next PR.

steplibrary/steplib.go
```diff
@@ -0,0 +1,34 @@
...
+	YMLPath, CodePath string
+}
+
+// New builds a Steplib. steplibURI is the steplib identity; inventoryURL is
+// the base URL the V2 inventory JSON is fetched from.
+func New(log stepman.Logger, steplibURI, inventoryURL string, fileManager fileutil.FileManager) *Steplib {
+	return &Steplib{
+		log:         log,
+		steplibURI:  steplibURI,
+		api:         NewHTTPAPI(inventoryURL, httpfetch.NewClient(log)),
```
godrei, line 31: do we really need two distinct clients for api and fetcher?

internal/httpfetch/httpfetch.go
```diff
@@ -0,0 +1,166 @@
...
+// path/hash are empty; on success the caller owns cleanup.
+func (c *client) fetchToTemp(ctx context.Context, dir, url string) (path string, hash string, err error) {
+	// Place the temp file alongside destPath so the final rename stays on
+	// one filesystem (cross-filesystem renames fail on most kernels).
+	tmp, err := os.CreateTemp(dir, "download-*.tmp")
+	if err != nil {
+		return "", "", fmt.Errorf("create temp file in %s: %w", dir, err)
+	}
+	defer func() {
+		if closeErr := tmp.Close(); closeErr != nil {
```
godrei, line 155: worth adding a comment if this is intentional: a successful download, but failed close results in a hard failure

steplibrary/resolve.go
```diff
@@ -0,0 +1,153 @@
...
+	var err error
+	if stepID == "" {
+		err = errors.New("missing required input: step id")
+	}
+
+	var allSteps []string
+	if err == nil {
+		allSteps, err = s.api.GetAllStepIDs(ctx)
+		if err != nil {
+			err = fmt.Errorf("fetching avaialble step IDs: %w", err)
```
godrei: typo: available

activator/result/result.go
```diff
@@ -0,0 +1,33 @@
...
+	StepYMLPath string
+
+	ActivationType ActivationType
+
+	// ExecutablePath is a local path to the main entrypoint of the step, ready for execution.
+	// This can be an empty string if:
+	// - step was activated from a git reference (we checked out the source dir directly)
+	// - step was activated from a local path (we copied the source dir directly)
+	// - step was activated from a steplib reference, but step.yml has no entry for pre-compiled binaries (we fallback to source checkout)
+	// - step was activated from a stpelib reference, but step.yml has no pre-compiled binary for the current OS+arch combo (we fallback to source checkout)
```
godrei: typo: steplib

godrei (commented):
Sequential if err == nil blocks are harder to read, use the standard Go style instead.

activator/result/result.go
lpusok: This file is moved from https://github.com/bitrise-io/stepman/blob/master/activator/activator.go, so already exists. (now diff shows file move)

internal/httpfetch/httpfetch.go
lpusok: The issue is that if rename suceeds then there is nothing to cleanup.
defer func(){  _ := os.Remove(tmpPath)}
this works, if we are ok with ignoring error (as cannot remove already renamed file)

steplibrary/steplibrary.go
lpusok: It already exists, changed it so file move is visible in diff.
+ the exhauststruct linter

internal/httpfetch/httpfetch.go
lpusok: Trimmed to a single line in 67acb88 — dropped the 'who satisfies it' note and the package names.

internal/httpfetch/httpfetch.go
lpusok: Good call — `Get` now reads a bounded 1KB snippet of the body into the non-2xx error (read error joined in, body still closed). 67acb88

internal/httpfetch/httpfetch.go
lpusok: Done — folded into a shared `download()` helper that uses `err`. 67acb88

internal/httpfetch/httpfetch.go
lpusok: Switched to a single best-effort `defer func(){ _ = os.Remove(tmpPath) }()` (harmless once the rename succeeds), per your suggestion. 67acb88

internal/httpfetch/httpfetch.go
lpusok: Extracted `download(ctx, destPath, url, expectedHash)` — empty hash skips verification, so `Download`/`DownloadWithHash` are now one-liners. 67acb88

internal/httpfetch/httpfetch.go
lpusok: Added a comment: a failed `Close` is an intentional hard failure (the final write may not have flushed). 67acb88

internal/httpfetch/httpfetch.go
lpusok: simplifed

internal/httpfetch/httpfetch.go
lpusok: extracted download()

steplibrary/steplibrary.go
lpusok: Done in 5cb0298 — `Activate` now uses standard `if err != nil { return … }` early returns right after each call.

steplibrary/resolve.go
lpusok: Reworked in 5cb0298 — replaced the `var err error`/`if err == nil` accumulator with standard early returns; the version switch moved into a `resolveVersion` helper.

steplibrary/resolve.go
lpusok: Changed in 5cb0298 — `resolveMinorLocked` now returns the parse error instead of skipping. Matches the `stepman` package, which already errors on `version.NewVersion` failures, and v2 `versions.json` holds only valid semver.

steplibrary/http_api.go
lpusok: Resolved by the PR1.5 merge — `http_api.go` now uses the exported `steplibindex` path helpers (`StepIDsPath().URL()`, `StepJSONPath(id, ver).URL()`, …).

steplibrary/resolve.go
lpusok: Fixed in 5cb0298 ("avaialble" → "available").

activator/result/result.go
lpusok: Fixed ("stpelib" → "steplib").

steplibrary/steplib.go
lpusok: yes, used in next PR

activator/activator_shim.go
```diff
@@ -0,0 +1,19 @@
+package activator
+
+import "github.com/bitrise-io/stepman/activator/result"
+
+// Backwards-compatible aliases for the activation result types. The canonical
```
godrei: Do we really need this file? If I get it correctly this just supports avoiding this migration in the CLI:
- activator.ActivationType -> result.ActivationType
- ...
This seems to be an easy migration for me, why do we want to avoid this?

activator/git_ref.go
```diff
@@ -50,9 +51,10 @@ even if the repository is open source!`)
 	}
 
 	return ActivatedStep{
-		StepYMLPath:     stepYMLPath,
+		StepInfo:         models.StepInfoModel{},
```
godrei: Is StepInfo always an empty struct when activating a step?

steplibrary/http_api.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (h *HTTPAPI) GetLatestStepVersions(ctx context.Context, id string) (steplibindex.LatestPointer, error) {
+	//nolint:exhaustruct // server JSON dictates which fields are populated
+	var out steplibindex.LatestPointer
+	p, err := steplibindex.LatestPointerPath(id)
+	if err != nil {
+		return out, err
+	}
+	err = h.fetchJSON(ctx, p.URL(), &out)
+	return out, err
```
godrei: NIT: without the explicit error handling and returning an empty struct in case of error, it took me a bit more time to understand whether we return some partial data or not in case of error.

steplibrary/http_api.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// GetStepModel fetches the V2 step manifest (`steps/<id>/<v>/step.json`,
+// which serializes models.StepModel) and returns the decoded model.
+func (h *HTTPAPI) GetStepModel(ctx context.Context, step ResolvedStepVersion) (models.StepModel, error) {
+	//nolint:exhaustruct // server JSON dictates which fields are populated
```
godrei: What does this lint do? Why do we need to skip it?

steplibrary/resolve_version.go
```diff
@@ -0,0 +1,152 @@
...
+
+	allSteps, err := s.api.GetAllStepIDs(ctx)
+	if err != nil {
+		return models.StepInfoModel{}, ResolvedStepVersion{}, fmt.Errorf("fetching available step IDs: %w", err)
+	}
+	if !slices.Contains(allSteps, stepID) {
+		return models.StepInfoModel{}, ResolvedStepVersion{}, fmt.Errorf("%s steplib does not contain %s step", s.steplibURI, stepID)
+	}
+
+	versionConstraint, err := models.ParseRequiredVersion(version)
```
godrei, line 20: NIT: this verification could be moved up next to the stepID verification

steplibrary/steplibrary.go
```diff
@@ -0,0 +1,68 @@
...
+
+type Steplib struct {
+	log stepman.Logger
+	// steplibURI is the steplib *identity* — the URI the user references in
+	// bitrise.yml (e.g. the official git URL). It is reported as
+	// StepInfoModel.Library.
+	steplibURI  string
+	api         API
+	fileManager fileutil.FileManager
+	fetcher     httpfetch.Client
```
godrei: Is it used?

steplibrary/steplibrary.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	stepModel, err := s.api.GetStepModel(ctx, resolved)
+	if err != nil {
+		return result.ActivatedStep{}, err
+	}
+
+	stepYML, err := yaml.Marshal(stepModel)
+	if err != nil {
+		return result.ActivatedStep{}, fmt.Errorf("marshal step model to YAML: %w", err)
```
godrei: Is this error intentionally wrapped and others are not?

activator/path_ref.go
godrei: Debugf is format, args... so this prints with %!(EXTRA …)

lpusok: Good catch — fixed in a338573: now uses `%s` format verbs instead of passing the values as extra args.

steplibrary/steplibrary.go
lpusok: No, it was unused — removed the `fetcher` field in a338573.

steplibrary/http_api.go
lpusok: The directive was unnecessary: this is a `var` declaration (not a composite literal), and exhaustruct's `allow-empty: true` permits empty values anyway. Removed in 0b9a940. The getter was also reworked in 948d1ea (see below).

steplibrary/http_api.go
lpusok: Agreed. Reworked all single-fetch getters in 948d1ea to return an empty struct explicitly on every error path (matching the slice getters), so we never surface a partially-decoded value on error.

steplibrary/resolve_version.go
lpusok: Done in 39ff1d1 — moved the version-constraint validation up next to the step-id check, so both inputs are validated before any network call.

steplibrary/steplibrary.go
lpusok: Made it consistent in 3fba068 — all four `Activate` error returns are now wrapped with a short breadcrumb (`resolve step version` / `fetch step definition` / `marshal step model to YAML` / `write step.yml`).

activator/activator_shim.go
lpusok: Agreed — removed the shim in 8a3138b. Moved the result types (`ActivatedStep`, `ActivationType` + the constants) into the `steplibrary` package, which is their canonical owner, and updated the activator callers to reference `steplibrary.ActivatedStep` / `steplibrary.ActivationType*` directly. The `activator/result` package is gone; the CLI does the same one-line import migration on its side.

activator/git_ref.go
lpusok: Not anymore — as of 57848f7, git-ref and path-ref activation populate `StepInfo` from the local `step.yml`, matching the steplib path. Both reuse stepman's existing builders: path activation calls `QueryStepInfoFromPath`, and for git I extracted `QueryStepInfoFromGitStepDir` from `QueryStepInfoFromGit` so the already-cloned dir reuses the same builder instead of re-cloning. The populated fields (`Library` = `git`/`path`, `ID`, `Version`, `Step`, `DefinitionPth`) match the conventions the CLI's `newStepInfoPtr` already uses, so this is a step toward retiring the `stepInfoPtr` workaround.

godrei (approved):

PR #381, opened 2026-06-03
test(indexgen): V2 inventory validator + wire-format / determinism guarantees

gergely-sallai:
## What & why

**PR 1.5 of the V2 steplib series** — follows PR 1 (V2 wire types + inventory generator, now merged to master). It adds a structural validator that runs on every generation, makes `steplibindex` the single source of truth for the inventory layout, and locks the wire-format and determinism contracts in tests.

In one sentence: **make invalid V2 trees structurally impossible to publish, and lock the layout + wire-format contracts in tests.**

### 1. `Validate` — `steplibrary/indexgen/validate.go` (new)

Walks a V2 inventory tree (rooted at the dir containing `v2/`) and returns the structural violations (`[]ValidationError`, sorted by `(Path, Msg)` for deterministic output):

| Tier | Asserts |
|---|---|
| Presence | `v2/meta.json`, `v2/index/step_ids.json` exist + parse. |
| Schema | `format_version` matches `steplibindex.FormatVersion`; `updated_at` non-zero; `step_ids` sorted + duplicate-free. |
| Cross-file | `latest.json#latest` is in `versions.json`; each `latest_by_major[N]` is in `versions.json` and prefixed `N.`; `step_id` in `latest.json`/`versions.json` equals the id. |
| Per-version | every version string in `versions.json` has a parseable `v2/steps/<id>/<v>/step.json` with non-empty `source.git` + `source.commit`. |
| Step info | `step-info.json` is mandatory (flagged if missing); every `asset_urls` entry must be a step-relative path — absolute URLs, absolute paths, and `../` traversal are flagged — that resolves to a real file on disk. |
| Stale | any file under `v2/steps/` or `v2/index/` not consumed above is flagged; a missing or unreadable expected root dir is itself a violation. |

### 2. Publish gate

The generator validates the fully-staged tree **before** its atomic publish rename; on any violation it returns an error and leaves any existing inventory at `outputDir` untouched. On success the published inventory is guaranteed to pass `Validate`. (Keeps PR 1's `Generate` API — no Stage/Publish split.)

### 3. `steplibindex` is the single source of truth for the layout — `paths.go` (new)

Every V2 file location is one constructor returning a `Path`, used by **both** the generator (write) and the validator/reader (read), so the two can't drift:
- `Path` exposes both forms, built once from the same segment list: `FS()` — slash path under the inventory root, for `fs.FS` lookups and the generator's writer; `URL()` — absolute, with `url.PathEscape` on the dynamic id/version segments, for the HTTP reader (PR 2).
- Dynamic segments are **validated** as safe single path components: a constructor returns an error if a segment is empty, `.`/`..`, or contains a path separator or NUL — so a crafted id can't read or write outside the step's subtree.
- The generator's writer is rooted at the inventory root and emits every path through these helpers (no hand-built literals), matching what the validator reads.

## Test guarantees

| File | Locks in |
|---|---|
| `validate_test.go` | each check fires on its specific corruption (missing meta, bad format_version, unsorted step_ids, latest → missing version, wrong `latest_by_major` major, missing step-info, absolute/traversing `asset_urls`, stale files, …); a freshly generated tree passes clean. |
| `roundtrip_integration_test.go` _(build tag `integration`)_ | every `step.yml` in the real default steplib round-trips to a `models.StepModel` whose JSON is semantically equal to the generated `step.json` (`assert.JSONEq`) — the V2 wire format is lossless against the runtime struct. |
| `determinism_test.go` | two generations with identical inputs produce byte-identical trees (per-file SHA-256). |
| `paths_test.go` | the FS and URL form of every path; `url.PathEscape` of dynamic segments (space, `?`, …) and rejection of unsafe segments (empty, `.`/`..`, separators). |

## Notable decisions

- **Validation is unconditional** — no skip flag. A failure is a bug worth blocking on.
- **`steplibindex` path constructors validate dynamic segments** — an unsafe id/version is rejected with an error, never spliced into a filesystem or URL path.
- **No `latest_versions.json` catalog and no `has_executable`** — dropped in the schema; the validator and paths reflect the `v2/index` layout with `versions`/`asset_urls` as string lists.
- **Round-trip compares via `assert.JSONEq`, not `reflect.DeepEqual`** — nil vs empty slice serialize identically under the wire contract.
- **Round-trip is a build-tagged integration test** (`//go:build integration`, runs against the real steplib, wired into the CI integration step) — excluded from a normal `go test ./...`. `generate_local_test.go` stays a `RUN_STEPLIB_GEN`-gated manual tool.

### DIFF
--- a/steplibrary/indexgen/determinism_test.go
+++ b/steplibrary/indexgen/determinism_test.go
@@ -5,10 +5,8 @@ import (
 	"encoding/hex"
 	"io/fs"
 	"os"
-	"path/filepath"
 	"testing"
 
-	"github.com/bitrise-io/stepman/internal/specfixtures"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -25,18 +23,10 @@ import (
 // Comparison is by per-file SHA-256 keyed on the output-relative path
 // (covering the whole v2/ tree); a failure prints which path's hash differs.
 func TestGenerator_deterministic_output(t *testing.T) {
-	opts := Options{GeneratedAt: fixedTime, SteplibCommitSHA: "deadbeefcafef00d"}
-
-	out1 := t.TempDir()
-	_, err := generateFromSteplibClone(specfixtures.SteplibClone(), out1, opts, testLogger{t})
-	require.NoError(t, err, "first generation")
-
-	out2 := t.TempDir()
-	_, err = generateFromSteplibClone(specfixtures.SteplibClone(), out2, opts, testLogger{t})
-	require.NoError(t, err, "second generation")
-
-	hashes1 := hashAllFiles(t, out1)
-	hashes2 := hashAllFiles(t, out2)
+	// Two runs of the same helper = identical fixture, Options, and logger; any
+	// output difference is therefore non-determinism in the generator.
+	hashes1 := hashAllFiles(t, runGenerateFromSteplibClone(t))
+	hashes2 := hashAllFiles(t, runGenerateFromSteplibClone(t))
 
 	require.NotEmpty(t, hashes1, "first generation produced no files")
 	assert.Equal(t, hashes1, hashes2,
@@ -56,7 +46,7 @@ func hashAllFiles(t *testing.T, dir string) map[string]string {
 		if d.IsDir() {
 			return nil
 		}
-		bytes, err := os.ReadFile(filepath.Join(dir, p))
+		bytes, err := fs.ReadFile(dirFS, p)
 		if err != nil {
 			return err
 		}
--- a/steplibrary/indexgen/determinism_test.go
+++ b/steplibrary/indexgen/determinism_test.go
@@ -1,0 +1,68 @@
+package indexgen
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/bitrise-io/stepman/internal/specfixtures"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGenerator_deterministic_output asserts that two generations with
+// identical inputs and Options produce byte-identical output trees.
+//
+// This makes the regen-a-sample-tree workflow's implicit assumption an
+// asserted contract. It also catches any future non-determinism (map
+// iteration order leaking into output, time.Now() injected somewhere other
+// than Options.GeneratedAt, etc.) before it confuses anyone diffing a
+// regenerated sample tree.
+//
+// Comparison is by per-file SHA-256 keyed on the output-relative path
+// (covering the whole v2/ tree); a failure prints which path's hash differs.
+func TestGenerator_deterministic_output(t *testing.T) {
+	opts := Options{GeneratedAt: fixedTime, SteplibCommitSHA: "deadbeefcafef00d"}
+
+	out1 := t.TempDir()
+	_, err := generateFromSteplibClone(specfixtures.SteplibClone(), out1, opts, testLogger{t})
+	require.NoError(t, err, "first generation")
+
+	out2 := t.TempDir()
+	_, err = generateFromSteplibClone(specfixtures.SteplibClone(), out2, opts, testLogger{t})
+	require.NoError(t, err, "second generation")
+
+	hashes1 := hashAllFiles(t, out1)
+	hashes2 := hashAllFiles(t, out2)
+
+	require.NotEmpty(t, hashes1, "first generation produced no files")
+	assert.Equal(t, hashes1, hashes2,
+		"two generations with identical Options must produce byte-identical trees")
+}
+
+// hashAllFiles walks dir and returns a map from output-relative path to the
+// SHA-256 hex digest of that file's contents. Directories are skipped.
+func hashAllFiles(t *testing.T, dir string) map[string]string {
+	t.Helper()
+	dirFS := os.DirFS(dir)
+	out := map[string]string{}
+	require.NoError(t, fs.WalkDir(dirFS, ".", func(p string, d fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
+			return walkErr
+		}
+		if d.IsDir() {
+			return nil
+		}
+		bytes, err := os.ReadFile(filepath.Join(dir, p))
+		if err != nil {
+			return err
+		}
+		sum := sha256.Sum256(bytes)
+		out[p] = hex.EncodeToString(sum[:])
+		return nil
+	}))
+	return out
+}
--- a/steplibrary/indexgen/indexgen.go
+++ b/steplibrary/indexgen/indexgen.go
@@ -197,11 +197,14 @@ func writeInventory(w *writer, inputFS fs.FS, steps []parsedStep, steplibYML mod
 }
 
 // publish atomically swaps the staged tree in for any existing one at outputDir.
-func publish(staging, outputDir string) error {
+func publish(stagingDir, outputDir string) error {
+	if _, err := os.Stat(stagingDir); err != nil {
+		return fmt.Errorf("staging dir (%s): %w", stagingDir, err)
+	}
 	if err := os.RemoveAll(outputDir); err != nil {
 		return fmt.Errorf("clear output dir %s: %w", outputDir, err)
 	}
-	if err := os.Rename(staging, outputDir); err != nil {
+	if err := os.Rename(stagingDir, outputDir); err != nil {
 		return fmt.Errorf("publish inventory to %s: %w", outputDir, err)
 	}
 	return nil
--- a/steplibrary/indexgen/indexgen.go
+++ b/steplibrary/indexgen/indexgen.go
@@ -133,9 +133,10 @@ func generateFromSteplibClone(inputFS fs.FS, outputDir string, opts Options, log
 		}
 	}()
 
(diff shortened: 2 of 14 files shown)


steplibrary/indexgen/determinism_test.go
```diff
@@ -0,0 +1,67 @@
...
+// this property; this test makes it an asserted contract. It also catches
+// any future non-determinism (map iteration order leaking into output,
+// time.Now() injected somewhere other than Options.GeneratedAt, etc.)
+// before it confuses anyone diffing a regenerated sample tree.
+//
+// Comparison is by per-file SHA-256 keyed on the inventory-relative path;
+// a failure prints which path's hash differs.
+func TestGenerator_deterministic_output(t *testing.T) {
+	inputFS := os.DirFS("testdata/input")
+	opts := Options{GeneratedAt: fixedTime, SteplibCommitSHA: "deadbeefcafef00d"}
```
godrei: I will open a cafe and name it based on this commit hash: dead beef cafe ☕️ .

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,280 @@
...
+	if _, err := fs.Stat(v.fs, p); err != nil {
+		// step-info.json is optional. If absent, nothing to do.
+		return
+	}
+	var info spec.StepInfo
+	if !v.readJSON(p, &info) {
+		return
+	}
+	for name, rel := range info.AssetURLs {
+		if strings.HasPrefix(rel, "http://") || strings.HasPrefix(rel, "https://") {
```
godrei: Are both the local and remote asset url cases possible?

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,280 @@
...
+//     per-file assertions would miss.
+//
+// The returned error (the second return value) is reserved for situations
+// where the walk itself cannot proceed (e.g., the filesystem is unreadable).
+// "The tree is malformed" is reported via the ValidationError slice, not
+// the error.
+func Validate(inventoryFS fs.FS) ([]ValidationError, error) {
+	v := &validator{fs: inventoryFS, seen: map[string]bool{}, errs: nil}
+	v.run()
+	return v.errs, nil
```
godrei: The second error param sees to be a premature optimization, that we might never will need.
I first read the consumer side code and was very curious about the two error return params, here it turned out we have only one in reality :( .

steplibrary/specgen/generator.go
```diff
@@ -143,11 +159,78 @@ func GenerateFromSteplibClone(inputFS fs.FS, outputDir string, opts Options, log
...
+// existing tree at outputDir is replaced wholesale. The rename consumes
+// stagingDir; callers should not attempt to clean it up after a successful
+// Publish (RemoveAll on the now-gone path is a no-op, so deferred cleanup
+// is safe).
+//
+// The atomicity guarantee depends on stagingDir and outputDir being on the
+// same filesystem, which StageFromSteplibClone ensures by creating the
+// staging dir as a sibling of outputDir.
+func Publish(stagingDir, outputDir string) error {
+	if err := os.RemoveAll(outputDir); err != nil {
```
godrei: It removed outputDir before validating if stagingDir exists, can it be an issue?

```
if _, err := os.Stat(stagingDir); err != nil {
      return fmt.Errorf("staging dir %s: %w", stagingDir, err)
  }
```
can prevent it

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+}
+
+// checkNoStaleFiles walks steps/ and spec/ once each and flags any file
+// the validator did not consume above. This catches "left-over files from a
+// previous generation" — e.g., a step that was later removed, or a stray
+// file from a generator bug.
+func (v *validator) checkNoStaleFiles() {
+	for _, root := range []string{spec.StepsRootFS, spec.SpecRootFS} {
+		_ = fs.WalkDir(v.fs, root, func(p string, d fs.DirEntry, err error) error {
```
godrei: Here we swallow walk errors is it intentional?

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+		v.consume(assetPath)
+	}
+}
+
+// checkNoStaleFiles walks steps/ and spec/ once each and flags any file
+// the validator did not consume above. This catches "left-over files from a
+// previous generation" — e.g., a step that was later removed, or a stray
+// file from a generator bug.
+func (v *validator) checkNoStaleFiles() {
```
godrei: Would it catch if an asset present on FS, but isn't in the step-info?

steplibrary/indexgen/validate.go
gergely-sallai: Nice find! This is always local (as in relative path). Claude tried to be defensive here, but as per the current spec an HTTP schemed remote (absolute) URL is a violation of the spec. :( And I didn't catch this.

Will correct it!
My idea is to expand on this to flag this case as an explicit violation of the spec, so it actually guards against something.
But of course I can drop the branching too, but that could let through a regression we could catch.

steplibrary/indexgen/validate.go
lpusok: Yes. Added test. Added fix for example if asset point to other step's dir, abs path or http path.

steplibrary/indexgen/validate.go
lpusok: Now we error out.

steplibrary/indexgen/validate.go
lpusok: decided to refactor to return violations explicitely (the walk error case added handling, but this is not reachable)

steplibrary/indexgen/validate.go
lpusok: removed

steplibrary/specgen/generator.go
lpusok: added check.

steplibrary/indexgen/indexgen.go
```diff
@@ -131,39 +133,86 @@ func generateFromSteplibClone(inputFS fs.FS, outputDir string, opts Options, log
 		}
 	}()
 
-	// Root the tree under the format-version dir (e.g. v2/) inside staging, so
-	// the published outputDir contains <version>/{meta.json,index,steps}.
-	w := &writer{outputDir: filepath.Join(staging, steplibindex.VersionDir()), fw: realFileWriter{}, fm: fileutil.NewFileManager(), fileCount: 0, byteCount: 0}
+	// The writer is rooted at the inventory root (staging); every path it gets
+	// comes from steplibindex, which roots files under the format-version dir
+	// (e.g. v2/), so the published outputDir contains <version>/{meta,index,steps}.
+	w := &writer{outputDir: staging, fw: realFileWriter{}, fm: fileutil.NewFileManager(), fileCount: 0, byteCount: 0}
```
godrei: NIT: A constructor could hide how to build this struct, highlight required params and ease understanding the code.

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,312 @@
...
+		return "escapes the step directory"
+	}
+	return ""
+}
+
+// staleFileViolations walks v2/steps and v2/index once each and returns a
+// violation for every file the checks above did not consume — left-over files
+// from a previous generation (a removed step), or a stray file from a generator
+// bug. It reads the accumulated seen set, so it must run after all other checks.
+func (v *validator) staleFileViolations() []ValidationError {
```
godrei, line 280: Would it make sense to check for stale files in the root dir too (`v2/`)?

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,312 @@
...
+	if step.Source.Commit == "" {
+		issues = append(issues, violationf(p, "missing source.commit"))
+	}
+	return issues
+}
+
+// checkStepInfo validates the step's step-info.json (at infoPath) and that each
+// asset_urls entry is a step-relative path resolving to a real file under stepDir.
+func (v *validator) checkStepInfo(infoPath, stepDir string) []ValidationError {
+	if _, err := fs.Stat(v.fs, infoPath); err != nil {
```
godrei, line 236: NIT: `v.readJSON` below also does the file existence validation + calls `consume`, so this block seems to be a duplicated work.

steplibrary/indexgen/validate.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	return issues
+}
+
+// validateAssetURL reports why rel is not a valid step-relative asset reference,
+// or "" if it is fine. The single rule — relative, scheme-less, and resolving
+// within stepDir — rejects absolute URLs, absolute paths, and parent-directory
+// traversal alike, so a bad entry can't slip through one gap while another is
+// guarded.
+func validateAssetURL(rel, stepDir string) string {
```
godrei: NIT: This is a bit strange that and empt string return value means: everything is OK.

steplibrary/indexgen/validate.go
lpusok: using v2 dir as root.

steplibrary/indexgen/indexgen.go
lpusok: added constructor.

steplibrary/indexgen/validate.go
lpusok: changed into checkAssetURL

godrei (approved):

PR #384, opened 2026-06-18
refactor(steplibrary): rename Steplib→Client, trim comments, map-driven fake

lpusok:
## Summary
Ancillary cleanup of the V2 read path (PR2, #377), with no behavior change. This is the **base of a 3-PR stack** that was split out of #378:
1. **(this) ancillary** — rename + comments + test fake
2. precompiled binary download (stacks on this)
3. source-resolution consolidation (stacks on 2)

## Changes
- Rename `steplibrary.Steplib` → `Client` (avoids the `steplibrary.Steplibrary` stutter); receivers become `c`.
- Trim verbose doc comments (`HTTPAPI`, `GetStepModel`, the version resolver) to what isn't obvious from the code.
- Replace the canned `FakeAPI` test fake with a single map-driven `fakeAPI` + `newFakeAPI()` constructor, so the downstream feature PRs can override fields without a second fake type.

## Test plan
- [x] `go test ./steplibrary/...`
- [x] `golangci-lint run ./steplibrary/...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #385, opened 2026-06-18
feat(steplibrary): precompiled binary download + shared precompiled handling

lpusok:
## Summary
Adds **precompiled-binary activation** to the V2 reader and consolidates precompiled handling so V1 (`activator/steplib`) and V2 (`steplibrary`) share one implementation. Stacks on #384 (now merged to `master`). Source resolution is **out of scope** here — it lands in #386.

## Changes
- **Move + share**: `activator/steplib/activate_executable.go` → `steplibrary/precompiled.go`, refactored into exported `ResolveExecutable` + `DownloadPrecompiled` (httpfetch-based). Storage-URL constants moved into `steplibrary`; `activator/steplib.ActivateStep` rewired onto the shared funcs. Replaces `retryablehttp` + manual `validateHash` with `httpfetch.DownloadWithHash`. The file is heavily rewritten in the move (git reports a low-similarity rename, ~17%), so GitHub shows it as add+delete — `git log --follow`/blame still track the move.
- **Wire V2**: `Client.Activate` prefers the precompiled binary for the current platform; when it isn't used the result is tagged `ActivationTypeSteplibSource` (actual source resolution is out of scope here and lands in #386).
- **Feature flag (both sites)**: `steplibrary.PrecompiledStepsEnabled()` (reads `BITRISE_EXPERIMENT_PRECOMPILED_STEPS`) gates precompiled use in **both** V1 `ActivateStep` and V2 `Client.Activate`; V1's local constant is removed.
- **`ResolveExecutable` returns the platform** `(executable, platform, ok)`, so both call sites get the `GOOS-GOARCH` key from one place (no inline `fmt.Sprintf`).
- **chmod `0o700`** on the downloaded binary (owner-only); error wording kept as `"set executable permission on file"`.
- **Standard error handling** in the download paths (`if err != nil { … }`, happy path outside).

## Tests
- `TestSteplib_Activate_Precompiled` — full `Client.Activate` precompiled path (flag, activation type, exec bit, step.yml, URL from `StorageURI`).
- `TestDownloadFromURLs` — restores the multi-URL fallback coverage lost with `activate_executable_test.go`, via a real `httpfetch.Client` + `httptest`: primary-only, 404→fallback, hash-mismatch→fallback, all-fail-lists-each, hash-mismatch→error.
- `TestBuildPrecompiledURLs` — URL building / http rejection.

## e2e (`bitrise_specv2.yml`)
Generation no longer calls the removed `bitrise steps generate-steplib`; it generates the V2 tree from this checkout by running the manual `indexgen.Generate` test, parameterized via env:
```
RUN_STEPLIB_GEN=1 STEPLIB_URI=… STEPLIB_OUT=… go test -run TestGenerate_local -v ./steplibrary/indexgen/
```
and sets `BITRISE_EXPERIMENT_PRECOMPILED_STEPS`. Note: exercising V2 end-to-end also needs bitrise's V2 activation routing (separate bitrise change) — building bitrise master currently uses the V1 path.

## Notes
- V1 keeps its experiment gate; V2 is now gated by the same flag (intentional parity).

## Test plan
- [x] `go test ./...`, `golangci-lint run ./steplibrary/... ./activator/...`


steplibrary/precompiled.go
```diff
@@ -0,0 +1,135 @@
...
+// to look up an entry in models.StepModel.Executables.
+func currentPlatform() string {
+	return fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)
+}
+
+// ResolveExecutable picks the precompiled binary for the current OS+arch from
+// the step model. It always returns the current platform key (for logging),
+// and ok is false when no precompiled binary is available, the platform isn't
+// covered, or the entry is missing storage_uri/hash.
+func ResolveExecutable(step models.StepModel) (executable models.Executable, platform string, ok bool) {
```
ofalvai: Is it even necessary to return `ok` from this function? Isn't it the same as `executable != nil`?

lpusok: Good call. `models.Executable` is a struct value (not nilable), so `ok` couldn't literally be replaced by `executable != nil` — but I changed `ResolveExecutable` to return `*models.Executable` (nil when no usable binary), which drops the `ok` bool and lets call sites guard on `executable != nil`. Done in cafbedf.

PR #387, opened 2026-06-19
Move activator types back to activator package

ofalvai:
PR #377 intentionally moved `ActivatedStep` / `ActivationType` into `steplibrary` as part of the V2 read path. That worked at the time, but when trying to establish a clean `activator` → `steplibrary` dependency direction, it immediately creates an import cycle — \`steplibrary\` owns the result type that \`activator\` needs to return.

This PR resolves that by separating concerns properly:
- `ActivatedStep` / `ActivationType` move back to `activator` where they belong — they describe the outcome of activation, not of a fetch
- `steplibrary.Activate` returns a minimal `ActivateResult` (only `StepInfo` + `StepYMLPath`) — the fetch/data layer exposes only what it actually produces
- `activator` can now import `steplibrary` without a cycle; `steplibrary` has no dependency on `activator`

### DIFF
--- a/steplibrary/steplibrary.go
+++ b/steplibrary/steplibrary.go
@@ -33,31 +33,28 @@ func New(log stepman.Logger, steplibURI, inventoryURL string, fileManager fileut
 	}
 }
 
-func (c *Client) Activate(ctx context.Context, stepID, version string, outputPaths ActivateOutputPaths) (ActivatedStep, error) {
+func (c *Client) Activate(ctx context.Context, stepID, version string, outputPaths ActivateOutputPaths) (ActivateResult, error) {
 	stepInfo, resolved, err := c.getStepVersionInfo(ctx, stepID, version)
 	if err != nil {
-		return ActivatedStep{}, fmt.Errorf("resolve step version: %w", err)
+		return ActivateResult{}, fmt.Errorf("resolve step version: %w", err)
 	}
 
 	stepModel, err := c.api.GetStepModel(ctx, resolved)
 	if err != nil {
-		return ActivatedStep{}, fmt.Errorf("fetch step definition: %w", err)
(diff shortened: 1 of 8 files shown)


steplibrary/steplibrary.go
```diff
@@ -33,31 +33,28 @@ func New(log stepman.Logger, steplibURI, inventoryURL string, fileManager fileut
 	}
 }
 
-func (c *Client) Activate(ctx context.Context, stepID, version string, outputPaths ActivateOutputPaths) (ActivatedStep, error) {
+func (c *Client) Activate(ctx context.Context, stepID, version string, outputPaths ActivateOutputPaths) (ActivateResult, error) {
```
ofalvai, line 36: Looks like this function is not the final implementation at this point, it doesn't fetch step source or binary yet.

gergely-sallai: We discussed this in detail on Slack.
However for the record. This needs a discussion along with the already waiting PRs on the v2 wiring.

This PR would eliminate a cyclic dependency between packages: activator and steplibrary. Which is very much desirable. However followup PRs of the original changes (now on master) behave a bit differently.

Both original `Activate(..)` and `steplibrary.Activate(..)` does end-to-end activation inside. Thus in theory they should rely on the same result class, but that causes the cyclic dependency...

We could either:
a) refactor the activation logic so that the metadata gathering is delegated to sub-packages like: `steplibrary`, but the actual activation (download, copy, whatever is needed) happens in an upper layer in a unified way.
b) accept that `Activate(..)` is end-to-end everywhere, but the result objects need a mapping on package boundaries.
c) Figure out another way of keeping the `ActivatedStep` as an output in all cases. Maybe move it to a third package?

lpusok (approved):

PR #389, opened 2026-06-19
go version update 1.25.11 

pgyula:
Bump go version to 1.25.11

Notes:
* TLS SHA-1 disabling: stepman actually makes HTTPS requests via hashicorp/go-retryablehttp (to GitHub, CDNs for step definitions). https://go.dev/doc/go1.25#cryptotlspkgcryptotls
* net/url IPv6 parsing (GO-2026-4601):  stepman genuinely calls url.Parse through retryablehttp. The fix makes parsing more correct, so well-formed URLs aren't affected --> tests confirm this.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bitrise-io/stepman
 
-go 1.25
+go 1.25.11
```
marcell-vida: Should we update version fixed in `.tool-versions` instead and keep this on non-exact version?

pgyula: i pushed a changed, also removed govulncheck form the WF

marcell-vida (approved):

ofalvai:
@pgyula please update the PR title, then it's good to merge

PR #391, opened 2026-06-23
Step 2437 wire in steplib api

lpusok:


activator/steplib/activate_executable.go
```diff
@@ -65,10 +63,6 @@ func activateStepExecutable(
 		return "", fmt.Errorf("set executable permission on file: %s", err)
 	}
 
-	if err := copyStepYML(stepLibURI, stepID, version, destinationStepYML); err != nil {
```
ofalvai, line 68: Are you sure this is not needed? I have vague memories that I tried to trace the old source-activation codepath and ended up with the conclusion that the `step.yml` needs to be in the dir that any activator returns. But I might be wrong, I tried to be extra cautious.

PR #392, opened 2026-06-24
[STEP-2437] Integrate steplib API into activation codepaths

ofalvai:
Integrates the steplib **V2 API** into the activation codepaths behind an opt-in flag, keeping the legacy (V1) local-steplib path as the default — plus the fixes, cleanups, and tests surfaced while wiring it up.

## Changes

**V2 API integration**
- Add the V2 API read path to `ActivateSteplibRefStep`, gated by `BITRISE_STEPLIB_API_ENABLE`. When enabled for the Bitrise steplib, steps are resolved/activated over the hosted steplib API instead of cloning the whole steplib; V1 stays the default. The API client is built inline when `shouldUseSteplibAPI` is true (nil otherwise); there is no endpoint override.
- `ActivateStep` now takes a `stepid.CanonicalID` and dispatches on a `libraryAPI` client (nil ⇒ legacy V1). It returns a `ResolvedStep{ExecPath, StepInfo}` — `ExecPath` is optional, set only for precompiled executable activation; `StepInfo` is populated on both paths.
- The V2 path resolves the step via `steplibrary.Client.FetchStepMetadata` (returns a `models.StepInfoModel`). `ActivateStep` writes `current_step.yml` itself, up front, from the resolved model.

**Legacy (V1) path**
- `ActivateStep`'s legacy branch resolves the step from the local steplib itself, via `stepman.QueryStepInfoFromLibrary` — the same lookup `prepareStepLibForActivation` uses — so both branches produce a full `StepInfoModel` and `ResolvedStep.StepInfo` is populated on the legacy path too (previously it came back empty there).
- `QueryStepInfoFromLibrary` resolves the constraint to a concrete version (e.g. `2.3 → 2.3.7`); using the raw constraint would hit the wrong cache path and break minor/major locks. This is safe on a no-match now that master's `latestMatchingStepVersion` returns not-found (STEP-2437, #399) instead of an empty, nil-`Source` step that panicked in source activation.

**Fixes / cleanups**
- Place `current_step.yml` once, up front in `ActivateStep`: copied from the local cache on the legacy path, written from the resolved model on the API path. (`FetchStepMetadata` no longer writes it — that removes the earlier collision where the source fallback re-copied it with an empty version.)
- Move `OriginalVersion` and the empty-title→step-ID default into `QueryStepInfoFromLibrary`, so every caller gets a complete `StepInfoModel` from one place; `prepareStepLibForActivation` no longer duplicates it.
- Drop the `NewResolvedStep` constructor in favor of plain struct literals — `exhaustruct` already enforces that every field is set. Make the source `ResolvedStep` return exhaustive; drop unused params (`resolveStepModel`'s `log`, `steplibrary.New`'s `steplibURI`).

**Tests**
- Native v1/v2 activation integration harness (`_tests/activation`, build tag `integration`) — in-process, real URLs; covers version forms (latest/exact/minor/major), precompiled vs source vs fallback, v1 cache cold/warm, offline mode, and error cases, logging v1 and v2 side by side. The v2 variants target the production inventory (no override).
- Unit tests for `ActivateSteplibRefStep` covering V1 resolution (exact + minor-lock), result wiring / `step.yml` placement, and the invalid-constraint / non-existent-version / non-existent-step error paths.
- Removed the superseded manual `bitrise_specv2.yml` e2e config.

## Testing

- `go build ./...`, `go vet ./...`, `golangci-lint` clean.
- Unit tests green: `models`, `activator`, `activator/steplib`, `steplibrary`, `stepman`, `cli`.
- Integration harness (`_tests/activation`) compiles under `-tags integration`; it needs network and is not part of CI.


activator/steplib/activate.go
```diff
@@ -22,38 +25,71 @@ var precompiledStepsDefaultStorageURLs = []string{
 	"https://storage.googleapis.com/bitrise-steplib-storage",
 }
 
-func ActivateStep(stepLibURI, id, version, destination, destinationStepYML string, log stepman.Logger, isOfflineMode bool) (string, error) {
-	stepCollection, err := stepman.ReadStepSpec(stepLibURI)
-	if err != nil {
-		return "", fmt.Errorf("failed to read %s steplib: %s", stepLibURI, err)
+type ResolvedStep struct {
+	ExecPath string
```
ofalvai, line 33: We should document that it's optional and not always has a non-empty value

lpusok: done.

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
 	}
 
-	step, version, err := queryStepMetadata(stepCollection, stepLibURI, id, version)
-	if err != nil {
-		return "", fmt.Errorf("failed to find step: %s", err)
+	// Place the step.yml at destinationStepYML once, up front: on the legacy
+	// path copy it from the local steplib cache; on the API path
+	// FetchStepMetadata has already written it there. Doing this before the
+	// precompiled/source split keeps it to a single callsite.
+	if libraryAPI == nil {
```
lpusok, line 56: v1 copies YML from the cloned steplib, so there is no obvious way around this.

steplibrary/steplibrary.go
```diff
@@ -17,13 +18,9 @@ type Client struct {
...
 
-type ActivateOutputPaths struct {
-	YMLPath, CodePath string
-}
-
-// New builds a Client. steplibURI is the steplib identity; inventoryURL is
-// the base URL the V2 inventory JSON is fetched from.
-func New(log stepman.Logger, steplibURI, inventoryURL string, fileManager fileutil.FileManager) *Client {
+// New builds a Client. inventoryURL is the base URL the V2 inventory JSON is
+// fetched from.
```
ofalvai:
```suggestion
// New builds a stepman.Client.
// inventoryURL: the base URL of the API where metadata is fetched from.
```

steplibrary/steplibrary.go
```diff
@@ -32,28 +29,29 @@ func New(log stepman.Logger, steplibURI, inventoryURL string, fileManager fileut
...
-		return ActivateResult{}, fmt.Errorf("write step.yml: %w", err)
+	if err := c.fileManager.WriteBytes(outputYMLPath, stepYML); err != nil {
+		return StepMetadata{}, fmt.Errorf("write step.yml: %w", err)
 	}
 
-	return ActivateResult{
+	return StepMetadata{
 		StepInfo:    stepInfo,
-		StepYMLPath: outputPaths.YMLPath,
+		StepYMLPath: outputYMLPath,
```
ofalvai: Can we eliminate this field? It's passed to this function as an argument, then here we return it to the caller. What if we just returned a `models.StepInfoModel` instance?

activator/steplib_ref.go
```diff
@@ -2,15 +2,25 @@ package activator
...
 	"github.com/bitrise-io/stepman/stepid"
+	"github.com/bitrise-io/stepman/steplibrary"
 	"github.com/bitrise-io/stepman/stepman"
 )
 
+const (
+	bitriseV1SteplibURL = "https://github.com/bitrise-io/bitrise-steplib.git"
+	// bitriseSteplibAPIURL is V2 Steplib API
+	bitriseSteplibAPIURL = "https://steplib.bitrise.io"
+	enableSteplibAPIEnv  = "BITRISE_EXPERIMENT_STEPLIB_API_ENABLE"
```
ofalvai: Suggestion: remove the `EXPERIMENT` part, it's going to stay with us for the foreseeable future.

activator/steplib_ref.go
```diff
@@ -2,15 +2,25 @@ package activator
...
 	"github.com/bitrise-io/stepman/activator/steplib"
 	"github.com/bitrise-io/stepman/models"
 	"github.com/bitrise-io/stepman/stepid"
+	"github.com/bitrise-io/stepman/steplibrary"
 	"github.com/bitrise-io/stepman/stepman"
 )
 
+const (
+	bitriseV1SteplibURL = "https://github.com/bitrise-io/bitrise-steplib.git"
+	// bitriseSteplibAPIURL is V2 Steplib API
```
ofalvai:
```suggestion
	// Base URL of steplib API
```

activator/steplib_ref.go
```diff
@@ -2,15 +2,25 @@ package activator
...
+	"github.com/bitrise-io/go-utils/v2/fileutil"
 	"github.com/bitrise-io/stepman/activator/steplib"
 	"github.com/bitrise-io/stepman/models"
 	"github.com/bitrise-io/stepman/stepid"
+	"github.com/bitrise-io/stepman/steplibrary"
 	"github.com/bitrise-io/stepman/stepman"
 )
 
+const (
+	bitriseV1SteplibURL = "https://github.com/bitrise-io/bitrise-steplib.git"
```
ofalvai: I think this naming is a bit misleading, this URL is not a "v1" URL, it's the official Bitrise steplib URL, used for checking which steplib the step belongs to.

activator/steplib_ref.go
```diff
@@ -47,6 +65,16 @@ func ActivateSteplibRefStep(
 	return activationResult, nil
 }
 
+// libraryAPIFactory builds a Steplib API client when the V2 read path
+// should be used, or returns nil to keep the legacy (V1) activation path.
+func libraryAPIFactory(steplibURI string, logger stepman.Logger) (client *steplibrary.Client) {
```
ofalvai: Can we rename this to something like `shouldUseSteplibAPI()` or similar? This is not a factory anymore, it just decides which codepath to use. I think it should just return a `boolean`, then instantiate the client one level higher

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
+	// ExecPath is optional: it holds the activated step executable only for
+	// precompiled executable activation, and is empty for source activation.
+	ExecPath string
+	StepInfo models.StepInfoModel
+}
+
+// NewResolvedStep builds a ResolvedStep. execPath is optional (empty for source activation).
+func NewResolvedStep(execPath string, stepInfo models.StepInfoModel) ResolvedStep {
+	return ResolvedStep{ExecPath: execPath, StepInfo: stepInfo}
+}
```
ofalvai: nit: is this needed? I don't have strong a strong opinion, but seems a bit pointless to me. We have the `exhausstruct` linter enabled anyway.

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
+
+func ActivateStep(id stepid.CanonicalID, destination, destinationStepYML string, log stepman.Logger, isOfflineMode bool, libraryAPI *steplibrary.Client) (ResolvedStep, error) {
+	var stepModel models.StepModel
+	var stepInfo models.StepInfoModel
```
ofalvai, line 38: bug: this doesn't get an assignment in the legacy codepath

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
+	if resolveErr != nil {
+		return NewResolvedStep("", models.StepInfoModel{}), resolveErr
 	}
 
-	step, version, err := queryStepMetadata(stepCollection, stepLibURI, id, version)
-	if err != nil {
-		return "", fmt.Errorf("failed to find step: %s", err)
+	// Place the step.yml at destinationStepYML once, up front: on the legacy
+	// path copy it from the local steplib cache; on the API path
+	// FetchStepMetadata has already written it there. Doing this before the
```
ofalvai: Is this comment still accurate? The copy call is wrapped in this `if`, so it's not called on the API codepath.

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
-		if err := copyStepYML(stepLibURI, id, version, destinationStepYML); err != nil {
-			return "", fmt.Errorf("copy step.yml: %s", err)
-		}
+		return NewResolvedStep(execPath, stepInfo), err
+	}
 
-		return execPath, err
+	// Fallback path to step source activation
+	// TODO: this is tied to the old stepman codepath because source activation needs a `stepCollection` object.
+	// Might be a good cleanup in a follow-up PR, maybe source activation can be made independent of `stepCollection`
```
ofalvai: I don't want to block the PR on this, just want to make sure we have a ticket for this cleanup, let's not forget it. It's a regression (source-activated steps don't get the speedup of the steplib API), I hope we can easily resolve this with a quick refactor.

activator/steplib/activate.go
```diff
@@ -22,38 +25,79 @@ var precompiledStepsDefaultStorageURLs = []string{
...
+	// TODO: this assumes that the step library spec is already up-to-date.
+	// This breaks when the new steplib API is NOT ENABLED and should be fixed in a follow-up PR. See steplib_ref.go.
```
ofalvai: Do you remember why we can't just call update here? I don't remember why I wrote this TODO :( It seems like a bad regression, CLI would resolve to slightly outdated step versions when the new API is not enabled (which is 100% of builds at the moment)

ofalvai: Or maybe this comment is no longer accurate because of https://github.com/bitrise-io/stepman/blob/953fbccfc2ef424de5d20c155fa1eae799586f79/activator/steplib_ref.go#L42?

steplibrary/steplibrary.go
ofalvai: Also, I didn't notice this earlier, but I think it's a bigger code smell: this code is called `FetchStepMetadata()` but it has a side-effect of writing the YML to disk too. I think that file write should be outside of `steplibrary`, it's a step activation concern.

activator/steplib/activate.go
lpusok: Followup: https://github.com/bitrise-io/stepman/pull/401
Let's do that quickly and leave the current TODO as is.

activator/steplib/activate.go
lpusok: Technically not a bug as for legacy path prepareSteplibForActivation returns it in ActivateSteplibRef. Meybe can make this clearer tough

activator/steplib_ref.go
lpusok: renamed.

activator/steplib_ref.go
lpusok: applied.

activator/steplib_ref.go
lpusok: Renamed.

activator/steplib_ref.go
lpusok: Renamed, moved constructor up.

activator/steplib/activate.go
lpusok: Removed constructor.

activator/steplib_ref.go
```diff
@@ -96,11 +120,6 @@ func prepareStepLibForActivation(
 		}
 	}
```
lpusok, line 119: Moved Title override (should be very rare), and originalveresion setting to QueryStepInfoFromLibrary, to be consistent for all callers.

activator/steplib/activate.go
lpusok: Updated, there are now two separate paths for legacy and API.

steplibrary/steplibrary.go
lpusok: Moved write out, removed StepMetadata.

steplibrary/steplibrary.go
lpusok: done.

activator/steplib/activate.go
ofalvai: I see it's the opposite problem than what the comment claimed, thanks for the rewording. And I agree it should be a follow-up PR.

lpusok:
@copilot resolve the merge conflicts in this pull request

gergely-sallai (approved):

PR #398, opened 2026-06-26
Move ActivatedStep

lpusok:
- **Moved ActivatedStep to setplib package**
- Return ActivatedStep from steplib.ActivateStep. This allows to move prepareSteplibForActivation one level down, into ActivateStep. It will make API integration easier: In the steplibrary.New we do not have seperate interfaces for "prepare" and "activate". Splitting it makes less sense as the second part always depends on the first.

next:
`//nolint:exhaustruct ` is only temporary, the plan is to merge prepareSteplibForActivation into steplib.ActivateStep.

### DIFF
--- a/activator/activate_ref_test.go
+++ b/activator/activate_ref_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/bitrise-io/go-utils/pathutil"
+	"github.com/bitrise-io/stepman/activator/steplib"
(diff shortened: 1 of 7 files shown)


ofalvai (changes requested):
It's a bit weird that the non-steplib activators now need to import the `steplib` package just for their return type. Let's talk about this, maybe there is a better way.

PR #401, opened 2026-07-08
[STEP-2437] V2-native step source activation (no SetupLibrary)

lpusok:
Follow-up to #392. Fixes step source download on the API activation path. It did not run *SetupLibrary*, so source fallback *activateStepSource* failed in a fresh environment (git steplib not cloned) with `no route found`. We do not add *SetupLibrary* as it would reintroduce the latency of git steplib clone.

Instead, we download the step source to a temp location. We reuse the step source download code from V1 (extracting where needed). The source "download locations" also stays the same, but now is returned by the API, as git steplib is not guaranteed to be cloned.

## Trade-offs (interim, follow-ups)

- **Caching:** API path does not cache step source archive locally. A dedicated V2 source cache is a follow-up.

### DIFF
--- a/activator/steplib/source.go
+++ b/activator/steplib/source.go
@@ -29,7 +29,7 @@ func activateStepSourceWithAPI(libraryAPI *steplibrary.Client, id, version strin
 		return fmt.Errorf("step %s@%s has no download location", id, version)
 	}
 
-	if err := stepman.DownloadStepArchive(destDir, locations, id, version, source.Commit, log); err != nil {
+	if err := stepman.DownloadStepSourceArchive(destDir, locations, id, version, source.Commit, log); err != nil {
 		return fmt.Errorf("download step source %s@%s: %s", id, version, err)
 	}
 	return nil
--- a/activator/steplib/source.go
+++ b/activator/steplib/source.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/bitrise-io/go-utils/pathutil"
 	"github.com/bitrise-io/stepman/models"
 	"github.com/bitrise-io/stepman/steplibrary"
 	"github.com/bitrise-io/stepman/stepman"
@@ -17,13 +16,7 @@ var ErrStepSourceNotCached = errors.New("step source not available in the local 
 
 // activateStepSourceWithAPI materializes id@version's source into destDir
 // without cloning a git steplib.
-func activateStepSourceWithAPI(libraryAPI *steplibrary.Client, uri, id, version string, source *models.StepSourceModel, destDir string, log stepman.Logger, isOfflineMode bool) error {
-	if reused, err := reuseCachedStepSource(uri, id, version, destDir); err != nil {
-		return err
-	} else if reused {
-		return nil
-	}
-
+func activateStepSourceWithAPI(libraryAPI *steplibrary.Client, id, version string, source *models.StepSourceModel, destDir string, log stepman.Logger, isOfflineMode bool) error {
 	if isOfflineMode {
 		return fmt.Errorf("%s@%s: %w", id, version, ErrStepSourceNotCached)
 	}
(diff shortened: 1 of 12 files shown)


steplibrary/steplibrary.go
```diff
@@ -40,3 +40,16 @@ func (c *Client) FetchStepMetadata(ctx context.Context, stepRef stepid.Canonical
 
 	return stepInfo, nil
 }
+
+// StepDownloadLocations resolves the concrete source download locations for
+// id@version from the inventory's base locations (meta.json download_locations),
+// preserving their order: a "zip" base becomes <base>/<id>/<version>/step.zip and
+// a "git" base becomes the step's own source git URL. For the bitrise steplib zip
+// comes first, so callers try the zip fast path before falling back to a git clone.
```
gergely-sallai: nit: This information in the doc is almost exactly the same as the comment on `BuildStepDownloadLocations`.

These kind of duplicates are very easy to drift over time, making it confusing.
I recommend aligning the two comments to only contain the information relevant to the actual part of the implementation.

E.g.: I think the
```
// preserving their order: a "zip" base becomes <base>/<id>/<version>/step.zip and
// a "git" base becomes the step's own source git URL. For the bitrise steplib zip
// comes first, so callers try the zip fast path before falling back to a git clone.
```
is an implementation detail of the `BuildStepDownloadLocations` function.

The only important thing here is that this function returns a priority order of source download URLs using information found on the Step Library API.

activator/steplib/source.go
```diff
@@ -0,0 +1,70 @@
...
+// mode when the step source is not in the local cache.
+var ErrStepSourceNotCached = errors.New("step source not available in the local cache and offline mode is set")
+
+// activateStepSourceV2 materializes id@version's source into destDir over the
+// steplib API, without setting up the local steplib. It reuses the V1 on-disk
+// cache when present; otherwise it downloads from the inventory's locations (zip
+// fast path, git fallback). The inventory locations are fetched only on the
+// download path, so offline mode and a cache hit do no network. Offline with no
+// cache returns ErrStepSourceNotCached.
+func activateStepSourceV2(libraryAPI *steplibrary.Client, uri, id, version string, source *models.StepSourceModel, destDir string, log stepman.Logger, isOfflineMode bool) error {
```
gergely-sallai: nit: Can we avoid baking in the `v2` into the function name?

- v2 is a very loaded term, we used it across multiple places, doesn't say much to fresh eyes here.
- this will age badly, both if we remove the 'legacy' route and if we must transition to v3

This function is almost the same as the original: `activateStepSource`, besides from where it works from, their aim is the same. They're in different files.
- Can we add a more descriptive name for `activateStepSourceV2`? -> activateStepSourceViaAPI
- Could they be in different packages with the same function name? (maybe overkill)

models/models_methods.go
```diff
@@ -317,29 +317,35 @@ func (collection StepCollectionModel) GetDownloadLocations(id, version string) (
...
+	}
+	return locations, nil
+}
+
+// BuildStepDownloadLocations resolves the concrete source download locations for
+// id@version from the steplib's base locations (steplib.yml / the V2 meta.json),
+// preserving their order: a "zip" base becomes <base>/<id>/<version>/step.zip and
+// a "git" base becomes the step's own source git URL (skipped when empty). An
+// unknown location type is an error.
+func BuildStepDownloadLocations(baseLocations []DownloadLocationModel, id, version, sourceGit string) ([]DownloadLocationModel, error) {
```
ofalvai: Can we have a more descriptive name for `sourceGit`? Something like `repoURL`

steplibrary/steplibrary.go
lpusok: Cleaned up comments.

activator/steplib/source.go
lpusok: Renamed to `activateStepSourceWithAPI`. Did not move packages, but like the idea, we could do this later when we add local caching.

models/models_methods.go
lpusok: Done.

activator/steplib/source.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/bitrise-io/stepman/models"
+	"github.com/bitrise-io/stepman/steplibrary"
+	"github.com/bitrise-io/stepman/stepman"
+)
+
+// activateStepSourceWithAPI materializes id@version's source into destDir
+// without cloning a git steplib.
+func activateStepSourceWithAPI(libraryAPI *steplibrary.Client, id, version string, source *models.StepSourceModel, destDir string, log stepman.Logger, isOfflineMode bool) error {
+	if isOfflineMode {
+		return fmt.Errorf("%s@%s: %w", id, version, errors.New("offline mode is not supported with Steplib API"))
```
ofalvai: nit: there is no need to use `errors.New()` and a `%s` here

ofalvai (dismissed):

gergely-sallai (approved):

PR #408, opened 2026-07-16
Add disk cache for precompiled step binaries

ofalvai:
### Why

There is no caching of (binary) step activations at the moment. Even in ephemeral CI environments, one workflow could reference the same step multiple times, fetching the binary for every execution.

### What

Building on top of #410, cache step binary downloads into `os.UserCacheDir()`.

Update: tested with integrating this PR into the CLI: https://github.com/bitrise-io/bitrise/pull/1354

### Decisions

- Responsibility of `httpfetch` vs `activator/steplib`: `httpfetch` should not be aware of our `sha256-xxxx` checksum prefix, it should work on explicitly SHA256 hashes. Handling of that `sha256-xxx` prefix is now the responsiblity of `activator/steplib`.
- No content-addressing. Step binaries are immutable per `(stepID, version, platform)`, so that triple is a sufficient cache key. There is no need to introduce content-addressing.
- No cache cleanup. Entries are written to `os.UserCacheDir()`, so they get periodically cleaned up. If this becomes an issue, we can add our own cleanup logic later.
- Hash is re-validated on every cache hit, not just at write time.

**Do not return cached executable path directly, copy into step destination dir**: Bitrise CLI runs in a wide variety of environments, including user-provided container images. We can't make assumptions about the cache directory's path and availability, especially when mounted across containers with different usernames, filesystems, etc. 

I looked into what it would take to safely use the per-user cache dir in all execution environments, and there are two issues that complicate things:
1. In the old Linux CI stack environments, Bitrise CLI runs inside a main container, and might execute steps in a "sidecar" container, all through the host's Docker daemon socket. User-level cache dir paths are not necessarily consistent across these 3 environments.
2. `$BITRISE_DOCKER_MOUNT_OVERRIDES` exists as an escape hatch, and the "step destination dir" is actually mounted across all 3 environments via this. We could add one more mount path, but this is already a bit hacky, and would require coordinating changes and rollouts across 3 systems (stepman, CLI, DEN)

### DIFF
--- a/activator/steplib/activate.go
+++ b/activator/steplib/activate.go
@@ -82,6 +82,8 @@ func ActivateStep(id stepid.CanonicalID, destination, destinationStepYML string,
 	}
 	stepModel := stepInfo.Step
 	version := stepInfo.Version
+	resolvedID := id
+	resolvedID.Version = version
 
 	// Place the step.yml at destinationStepYML once, up front.
 	if !useSteplibAPI {
@@ -94,7 +96,7 @@ func ActivateStep(id stepid.CanonicalID, destination, destinationStepYML string,
 		}
 	}
 
-	execPath, err := downloadPrecompiled(log, stepModel, id, destination, fetcher, opts)
+	execPath, err := downloadPrecompiled(log, stepModel, resolvedID, destination, fetcher, opts)
 	if execPath != "" {
 		return ResolvedStep{ExecPath: execPath, StepInfo: stepInfo}, err
 	}
@@ -127,7 +129,7 @@ func downloadPrecompiled(log stepman.Logger, step models.StepModel, id stepid.Ca
 		if ok && executableForPlatform.Hash != "" && executableForPlatform.StorageURI != "" {
 			log.Debugf("Downloading executable for %s", platform)
 			downloadStart := time.Now()
-			execPath, err := activateStepExecutable(context.Background(), fetcher, id.IDorURI, executableForPlatform, destination, log, opts.storageURLs())
+			execPath, err := activateStepExecutable(context.Background(), fetcher, id.IDorURI, id.Version, platform, executableForPlatform, destination, log, opts.storageURLs())
 			if err == nil {
 				log.Debugf("Downloaded executable in %s", time.Since(downloadStart).Round(time.Millisecond))
 
--- a/activator/steplib/activate.go
+++ b/activator/steplib/activate.go
@@ -129,7 +129,7 @@ func downloadPrecompiled(log stepman.Logger, step models.StepModel, id stepid.Ca
 		if ok && executableForPlatform.Hash != "" && executableForPlatform.StorageURI != "" {
 			log.Debugf("Downloading executable for %s", platform)
 			downloadStart := time.Now()
-			execPath, err := activateStepExecutable(context.Background(), fetcher, id.IDorURI, id.Version, platform, executableForPlatform, destination, log, opts.storageURLs())
+			execPath, err := activateStepExecutable(context.Background(), fetcher, id.SteplibSource, id.IDorURI, id.Version, platform, executableForPlatform, destination, log, opts.storageURLs())
 			if err == nil {
 				log.Debugf("Downloaded executable in %s", time.Since(downloadStart).Round(time.Millisecond))
 
--- a/activator/steplib/activate_executable.go
+++ b/activator/steplib/activate_executable.go
@@ -2,12 +2,17 @@ package steplib
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/bitrise-io/go-utils/v2/fileutil"
 	"github.com/bitrise-io/stepman/internal/httpfetch"
 	"github.com/bitrise-io/stepman/models"
 	"github.com/bitrise-io/stepman/stepman"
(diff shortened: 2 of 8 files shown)


activator/steplib/activate.go
```diff
@@ -52,6 +52,8 @@ func ActivateStep(id stepid.CanonicalID, destination, destinationStepYML string,
 	}
 	stepModel := stepInfo.Step
 	version := stepInfo.Version
+	resolvedID := id
```
ofalvai, line 55: We want to be careful not to use a major or major.minor version for caching concrete versions.

activator/steplib/activate_executable.go
```diff
@@ -16,22 +20,44 @@ import (
...
+	return cachePath, nil
+}
+
+func stepExecutableCachePath(stepID, version, platform string) (string, error) {
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+	base := filepath.Join(userCacheDir, "bitrise", "steps", "executables")
+	return filepath.Join(base, stepID, version, platform, stepID), nil
```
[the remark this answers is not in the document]
ofalvai: Fixed: https://github.com/bitrise-io/stepman/pull/408/commits/d5a96c85e0c43321cb93f2796fe90c887a60a9db

activator/steplib/activate_executable.go
```diff
@@ -16,22 +20,44 @@ import (
...
+		if rmErr := os.Remove(cachePath); rmErr != nil && !errors.Is(rmErr, fs.ErrNotExist) {
+			logger.Warnf("Failed to remove invalid cache entry %s: %s", cachePath, rmErr)
+		}
 	}
 
-	if err := os.Chmod(path, 0755); err != nil {
+	if err := downloadExecutable(ctx, fetcher, executable, cachePath, logger); err != nil {
+		return "", err
+	}
+	if err := os.Chmod(cachePath, 0755); err != nil {
```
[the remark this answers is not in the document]
ofalvai: I think the proper fix would involve changing the `httpfetch` package, so I'm working around it in this PR with this small change: https://github.com/bitrise-io/stepman/pull/408/changes/8fca3c5228e861ca7d086014c38b02d620d36482

FYI @lpusok

activator/steplib/activate_executable.go
```diff
@@ -16,22 +20,61 @@ import (
...
+	case err == nil:
+		// A concurrent activation, or one interrupted between the download's
+		// atomic publish and the chmod below, can leave a hash-valid cache
+		// entry that isn't executable yet. Repair the mode on every hit so
+		// such an entry doesn't stay broken forever.
+		if err := os.Chmod(cachePath, 0755); err != nil {
+			return "", fmt.Errorf("set executable permission on file: %s", err)
+		}
+		return cachePath, nil
+	case errors.Is(err, fs.ErrNotExist):
```
marcell-vida, line 51: Would it be helpful to print a custom message here for better debug experience, then fall through? e.g.
```go
logger.Warnf("Step executable was not found")
fallthrough
```

activator/steplib/activate_executable.go
```diff
@@ -16,22 +20,61 @@ import (
...
-	return path, nil
+	return cachePath, nil
+}
+
+func stepExecutableCachePath(steplibSource, stepID, version, platform string) (string, error) {
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+	base := filepath.Join(userCacheDir, "bitrise", "steps", "executables")
```
lpusok, line 97: we already use (.bitrise and .stepman dirs), maybe we can reuse one of them on a non-conflicting dubdir?

activator/steplib/activate_executable.go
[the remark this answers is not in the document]
lpusok: Hm yes, opening a pr for that.

activator/steplib/activate_executable.go
ofalvai: Those are outside the OS cache dir, so they are not cleaned up automatically (and also, I'd like to follow the [XDG spec](https://specifications.freedesktop.org/basedir/latest/))

lpusok, line 97: **Containerized steps will break: this path isn't visible inside the container.**

`ExecutablePath` isn't just opened by the CLI, it's passed to `docker exec` verbatim (`cli/run_util.go:682` -> `:743-744`), so it has to resolve *inside* the step container at the same absolute path.

That works today only because `destination` is `$TMPDIR/bitrise/step_src` and the mount overrides cover `/tmp` (see `integrationtests/docker/local_docker_test_environment/docker-compose.yml:8`). `os.UserCacheDir()` (`~/.cache` on Linux) isn't in that list, and the CLI supplies no default mounts (`cli/docker/container_manager.go:60-61`).

CI won't catch it either: the only precompiled-step integration test runs on the host, and every containerization test uses `script@1` (bash toolkit, no executable).

Worth checking the production `BITRISE_DOCKER_MOUNT_OVERRIDES` before merging. If `~/.cache` isn't mounted, options are to keep the cache as the source of truth but hand back a path under `destination` (copy or hardlink on hit), or move the cache somewhere already mounted.

activator/steplib/activate.go
```diff
@@ -64,7 +66,7 @@ func ActivateStep(id stepid.CanonicalID, destination, destinationStepYML string,
 		}
 	}
 
-	execPath, err := downloadPrecompiled(log, stepModel, id, destination, fetcher)
+	execPath, err := downloadPrecompiled(log, stepModel, resolvedID, fetcher)
```
lpusok: **Dropping `destination` here means `step_src` no longer exists at all for precompiled activations.**

`CleanupStepWorkDir` does a full `RemoveAll` of `step_src` before every activation (`bitrise/util.go:179-182`). Today the download recreates it as a side effect, because `httpfetch.download` does `MkdirAll(filepath.Dir(destPath))`. After this change nothing does, so `BITRISE_STEP_SOURCE_DIR` (`cli/run_util.go:510`) points at a directory that doesn't exist -- it used to exist and hold the binary.

Small blast radius for precompiled steps specifically, but it's a silent contract change: a step doing `cd $BITRISE_STEP_SOURCE_DIR` now fails.

activator/steplib/activate_executable.go
```diff
@@ -16,22 +20,61 @@ import (
...
+		// entry that isn't executable yet. Repair the mode on every hit so
+		// such an entry doesn't stay broken forever.
+		if err := os.Chmod(cachePath, 0755); err != nil {
+			return "", fmt.Errorf("set executable permission on file: %s", err)
+		}
+		return cachePath, nil
+	case errors.Is(err, fs.ErrNotExist):
+		// cache miss, fall through to download
+	default:
+		logger.Warnf("Cached step executable failed validation, re-downloading: %s", err)
```
lpusok, line 55: `validateHash` calls `parseExpectedHash` first, so an inventory entry with a non-`sha256-` prefix lands in this branch and logs "Cached step executable failed validation, re-downloading" -- when there is no cache entry and the fault is the steplib, not the cache. The `os.Remove` below is a no-op in that case too.

Parsing the hash once, before the cache lookup, keeps this branch about actual corruption.

activator/steplib/activate_executable.go
ofalvai: Thank you, added

activator/steplib/activate_executable.go
ofalvai: Good point, fixed: https://github.com/bitrise-io/stepman/compare/782952b0d1302607e983344718df3d0b474baed9..39a3d01036a8d779026b642fd0aed219e581b5ec

activator/steplib/activate_executable.go
```diff
@@ -16,23 +21,85 @@ import (
...
+	// A step ID is only unique within its StepLib, so the steplib source must be part
+	// of the cache key too, otherwise two libraries could publish different
+	// binaries under the same ID/version/platform and overwrite each other's
+	// cache entry.
+	// Hash the steplib string rather than using it as a path segment directly:
+	// steplibSource is typically a URL, and filepath.Join+Clean would collapse
+	// any ".." in it (path traversal), while characters like ":" in "https://"
+	// aren't valid in a Windows path segment.
+	sourceKey := sha256.Sum256([]byte(steplibSource))
+	return filepath.Join(base, hex.EncodeToString(sourceKey[:]), stepID, version, platform, stepID), nil
```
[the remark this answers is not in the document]
ofalvai: Step executables are only relevant for steplib step references, and steplib step IDs, versions, platforms are validated at publish time, they can't contain such characters.

activator/steplib/activate_executable.go
```diff
@@ -16,23 +21,85 @@ import (
...
+		}
+		if err := os.Chmod(cachePath, 0755); err != nil {
+			return "", fmt.Errorf("set executable permission on file: %s", err)
+		}
+	}
+
+	// Instead of returning the cache path, copy the executable to the destination dir.
+	// Bitrise CLI runs in a wide variety of environments, including user-provided container images.
+	// We can't make assumptions about the cache directory's path and availability, especially when
+	// mounted across containers with different usernames, filesystems, etc.
```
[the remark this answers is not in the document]
ofalvai: Updated description since then.

activator/steplib/activate_executable.go
```diff
@@ -16,23 +21,85 @@ import (
...
+	destPath := filepath.Join(destinationDir, stepID)
+	if err := os.MkdirAll(destinationDir, 0o755); err != nil {
+		return "", fmt.Errorf("create destination dir: %w", err)
+	}
+	if err := fileutil.NewFileManager().CopyFile(cachePath, destPath, &fileutil.CopyOptions{Overwrite: true}); err != nil {
+		return "", fmt.Errorf("copy cached executable to destination: %w", err)
+	}
```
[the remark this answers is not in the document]
ofalvai: Fixed: https://github.com/bitrise-io/stepman/compare/bc94eaed6695ed7b4f6a7a46d736fd37530cb24a..9d377a1ba27ff6e17ce7deffc168fc7bf0a60ca7

lpusok (approved):

marcell-vida (approved):

PR #405, opened 2026-07-10
refactor(steplib): route precompiled and step-zip downloads through internal/httpfetch

pgyula:

PR #410, opened 2026-07-17
refactor(stepman): route step-source zip downloads through internal/httpfetch

pgyula:
## Summary
Routes step-source zip downloads through `internal/httpfetch` instead of the vendored `command.DownloadAndUnZIP`, which does a bare `http.Get` with no retry and no atomicity. As per PR review [request](https://github.com/bitrise-io/stepman/pull/405)

## Changes
- `util.go`: zip downloads now go through the injected `httpfetch.Client` and extract via the new `internal/ziputil`. Drops the outer manual retry loop ( httpfetch already retries transient failures).
- `DownloadStep` now extracts into a temp staging dir and renames into place atomically, thus failed/partial extraction can no longer leave a half-populated dir that gets served as a cache hit.
- New `internal/ziputil`: a small stdlib-based replacement for `command.UnZIP`, which calls `log.Fatal` on close errors and has no zip-slip containment check.
- API change: `DownloadStep` and `DownloadStepSourceArchive` both gain a `fetcher httpfetch.Client` param. Only this repo calls them (CLI vendors but doesn't call), so blast radius is a module bump.

## Tests
- `util_test.go`: download success/failure, cache-dir staging contract
- `ziputil_test.go`: zip-slip rejection, permission preservation, corrupt archive
- `helpers_test.go`: fake fetcher now embeds a real `httpfetch.Client`

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...`
- [x] `golangci-lint run --config .golangci.yml` (v2.6.0) — 0 issues
- [x] `go test ./_tests/integration/...`
- [x] `go test -tags integration ./_tests/activation/...` — the real-steplib v1/v2 activation matrix, which exercises both the zip and git source paths through the new staging dir


Badlazzor (dismissed):

stepman/util.go
```diff
@@ -178,6 +176,22 @@ func DownloadStepSourceArchive(destDir string, downloadLocations []models.Downlo
...
+// extracts it into destDir.
+func downloadStepZip(log Logger, url, destDir string) error {
+	tmpDir, err := pathutil.NormalizedOSTempDirPath("stepman-step-zip")
+	if err != nil {
+		return fmt.Errorf("create temp dir: %w", err)
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
+
+	zipPath := filepath.Join(tmpDir, "step.zip")
+	if err := httpfetch.NewClient(log).Download(context.Background(), zipPath, url); err != nil {
```
gergely-sallai: Shouldn't `httpfetch.NewClient(log)` come from dependency injection instead?

I see this pattern everywhere in the code, and makes unit testing hard / impossible.

I see that we at the moment still want integration test as well because of the ZIP file, but what do you think?

pgyula: updated with the suggestion thanks!

gergely-sallai (dismissed):

stepman/util.go
```diff
@@ -186,6 +184,22 @@ func DownloadStepSourceArchive(destDir string, downloadLocations []models.Downlo
 	return errors.New("failed to download step")
 }
 
+// downloadStepZip fetches the step source archive at url into a temp file and
+// extracts it into destDir.
+func downloadStepZip(fetcher httpfetch.Client, url, destDir string) error {
```
lpusok: **The atomic rename here is wasted, and the thing that actually needs atomicity doesn't get it.**

This temp dir exists only to hold a file that's deleted three lines later, so `Download`'s temp-file-then-rename buys nothing — the destination is itself throwaway. Meanwhile `command.UnZIP` extracts *straight into* `destDir`, and on the legacy path `destDir` is the step cache, which is keyed by existence alone: `DownloadStep` above returns early if `stepPth` exists (`util.go:120-125`), and `activateStepSource` skips the download entirely when the cache dir is present (`activator/steplib/activate_source.go:26-42`).

`UnZIP` does `os.MkdirAll(dest, 0755)` as its very first action, so a failure on *any* entry (disk full, truncated archive, permissions) leaves the cache dir existing and half-populated. Today the following `git` location usually rescues it, because that branch starts with `os.RemoveAll(destDir)` — but if zip is the only location, or git also fails, the partial dir survives and the **next** run treats it as a valid cache hit and copies a broken step into the workflow.

This is pre-existing, not introduced here. But this PR makes it more reachable: the `retry.Times(2)` being removed used to re-download and re-extract over the same dir (`UnZIP` opens with `O_TRUNC`), which would often have repaired it on attempt 2.

**Recommendation:** put the staging where it earns its keep — in `DownloadStep`, which is the function that owns the cache and already guarantees `stepPth` does not exist:

```go
stepPth := GetStepCacheDirPath(route, id, version)
if exist, err := pathutil.IsPathExists(stepPth); err != nil {
	return err
} else if exist {
	return nil
}

// Materialize into a staging dir next to the cache dir and move it into place only
// once it is complete: the cache is keyed by the existence of stepPth alone, so a
// half-written stepPth would be served as a valid cache hit on the next run.
if err := os.MkdirAll(filepath.Dir(stepPth), 0o755); err != nil {
	return fmt.Errorf("create step cache dir: %w", err)
}
stagingDir, err := os.MkdirTemp(filepath.Dir(stepPth), ".staging-")
if err != nil {
	return fmt.Errorf("create staging dir: %w", err)
}
defer func() {
	if err := os.RemoveAll(stagingDir); err != nil {
		log.Warnf("Failed to remove staging dir %s: %s", stagingDir, err)
	}
}()

if err := DownloadStepSourceArchive(stagingDir, downloadLocations, id, version, commithash, log, fetcher); err != nil {
	return err
}
return os.Rename(stagingDir, stepPth)
```

`os.MkdirTemp` under `filepath.Dir(stepPth)` keeps the rename on one filesystem — the same reasoning `httpfetch.fetchToTemp` already documents for the file case.

Deliberately **not** suggesting this inside `downloadStepZip`: on the API path `destDir` is the activated-step dir the CLI creates before activation, and `rename` onto a non-empty directory is `ENOTEMPTY`. That path also doesn't need it — a fresh per-activation dir can't poison a cache.

With that in place, this helper's own temp dir can stay exactly as it is.

stepman/util.go
```diff
@@ -124,21 +126,17 @@ func DownloadStep(collectionURI string, collection models.StepCollectionModel, i
...
+func DownloadStepSourceArchive(destDir string, downloadLocations []models.DownloadLocationModel, id, version, commithash string, log Logger, fetcher httpfetch.Client) error {
 	for _, downloadLocation := range downloadLocations {
 		switch downloadLocation.Type {
 		case "zip":
-			err := retry.Times(2).Wait(3 * time.Second).Try(func(attempt uint) error {
-				return command.DownloadAndUnZIP(downloadLocation.Src, destDir)
-			})
-
-			if err != nil {
+			if err := downloadStepZip(fetcher, downloadLocation.Src, destDir); err != nil {
```
lpusok: On dropping the outer `retry.Times(2).Wait(3s)` — the latency numbers check out. `retryablehttp` defaults are `RetryMax=4` with 1s→30s exponential backoff (1+2+4+8 ≈ 15s), so stacking a 3-attempt outer loop on top really did give ~51s worst case. Removing one layer is the right call.

Worth being explicit in the description that the two layers weren't equivalent, though. `retryablehttp` retries at **request** level only: if the response starts fine and then dies mid-body (connection reset, truncated response), `fetchToTemp`'s `io.Copy` returns an error and nothing retries it. Same for a `UnZIP` failure on a corrupt archive, which is now outside any retry at all. The old outer loop covered both, because it wrapped download-and-extract as a unit. Step source zips are small so the window is narrow, but it's a coverage change rather than pure redundancy removal — and it's the reason the partial-extract issue in my other comment gets easier to hit.

The other half of the same change is a straight win and worth stating too: `retryablehttp` does **not** retry 4xx (except 429), so a 404/403 download location now fails immediately instead of burning 3 attempts and 6s of sleep before falling through to the git location.

stepman/util.go
```diff
@@ -186,6 +184,22 @@ func DownloadStepSourceArchive(destDir string, downloadLocations []models.Downlo
...
+	if err != nil {
+		return fmt.Errorf("create temp dir: %w", err)
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
+
+	zipPath := filepath.Join(tmpDir, "step.zip")
+	if err := fetcher.Download(context.Background(), zipPath, url); err != nil {
+		return fmt.Errorf("download step zip: %w", err)
+	}
+	return command.UnZIP(zipPath, destDir)
```
lpusok: **Suggest `go-utils/v2/ziputil` instead of `command.UnZIP`.**

`command.UnZIP` keeps two of the problems this PR is removing from `DownloadAndUnZIP`:

- Three `log.Fatal` calls in its deferred `Close` handlers (`go-utils@v1.0.13/command/zip.go:23,39,56`). That's `os.Exit(1)` from inside a library — it would kill the bitrise CLI mid-run, skipping the caller's error handling and any cleanup.
- `filepath.Join(dest, f.Name)` with no containment check, i.e. Zip Slip. For the canonical steplib the archive is served from Bitrise's own CDN, but `DownloadStepSourceArchive` also serves custom/self-hosted steplibs, where `download_locations` is whatever that steplib's own spec.json says.

`github.com/bitrise-io/go-utils/v2/ziputil` addresses both: stdlib `archive/zip`, no `log.Fatal`/`os.Exit` anywhere, rejects path-traversal entries, and it restores permissions and symlinks (v1's drops symlinks).

```go
zipManager := ziputil.NewZipManager(pathutil.NewPathChecker())
return zipManager.UnZip(zipPath, destDir)
```

Two caveats to weigh:

- `ziputil` isn't in a tagged release yet — it landed on go-utils/v2 master after `v2.0.0-alpha.36`, so this needs a pseudo-version bump or an `alpha.37` tag. Not a new dependency at least: we already use `go-utils/v2` (`alpha.35`, for `fileutil` in `activator/steplib/activate.go`).
- Its traversal check rejects any entry whose name *contains* `..` as a substring, not just as a path component — so an archive with a file literally named `foo..bar` would be refused. Documented upstream as deliberate (it's what makes CodeQL recognise the sanitizer). Very unlikely in step sources, but it is a narrowing.

If the untagged dependency is the blocker, the fallback is ~20 lines of stdlib locally, which gets the same two fixes without waiting on a release.

stepman/util.go
```diff
@@ -186,6 +184,22 @@ func DownloadStepSourceArchive(destDir string, downloadLocations []models.Downlo
...
 }
 
+// downloadStepZip fetches the step source archive at url into a temp file and
+// extracts it into destDir.
+func downloadStepZip(fetcher httpfetch.Client, url, destDir string) error {
+	tmpDir, err := pathutil.NormalizedOSTempDirPath("stepman-step-zip")
+	if err != nil {
+		return fmt.Errorf("create temp dir: %w", err)
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
```
lpusok: Nit, but it's a convention we hold in this repo: we don't discard errors with `_ =`. The instances in `internal/httpfetch/httpfetch.go` are the narrow exception, and each one carries a comment saying why — e.g. `httpfetch.go:129-131`:

```go
// Best-effort cleanup: on a verify/rename failure this removes the temp
// file; after a successful rename tmpPath is gone, so Remove fails harmlessly.
defer func() { _ = os.Remove(tmpPath) }()
```

and `:159-161`, where the comment spells out that the original error is what matters. A bare `_ =` with no such justification should just be handled — and here it's worth handling for real, since failing to clean up a temp dir is something you'd want to see in a log:

```go
defer func() {
	if err := os.RemoveAll(tmpDir); err != nil {
		log.Warnf("Failed to remove temp dir %s: %s", tmpDir, err)
	}
}()
```

That means threading `log` into `downloadStepZip`, which it arguably wants anyway.

lpusok (commented):
Reviewed the zip download/unzip path. The helper itself is about as small as it can get — `archive/zip` needs `io.ReaderAt`+size, so the archive has to be fully materialized and the temp file is unavoidable. Four comments inline: one correctness issue with a recommended fix, one note on what the retry change does and doesn't cover, one dependency suggestion, and one convention nit.

Separately, worth fixing in the PR body: it says the change touches only `stepman/util.go` and `util_test.go`, but it's 9 files and it changes two *exported* signatures (`DownloadStep`, `DownloadStepSourceArchive`). I checked — the Bitrise CLI vendors both and calls neither, so the practical blast radius is this repo plus the usual module bump, but that belongs in the description since exported changes are consumer-visible. Also stale: the "opening as a draft, hold until #404 lands" note — #404 is merged (`1509b9e`) and this is no longer a draft.

activator/steplib/helpers_test.go
```diff
@@ -157,7 +165,9 @@ type fakeExecutableFetcher struct {
 var _ httpfetch.Client = (*fakeExecutableFetcher)(nil)
 
 func (fakeExecutableFetcher) Get(context.Context, string) (io.ReadCloser, error) { return nil, nil }
-func (fakeExecutableFetcher) Download(context.Context, string, string) error     { return nil }
+func (fakeExecutableFetcher) Download(ctx context.Context, destPath, url string) error {
+	return httpfetch.NewClient(discardLogger{}).Download(ctx, destPath, url)
```
lpusok: Two things to fix here, and I think the test can be fixed properly rather than patched.

`discardLogger` throws the retry adapter's output away, which is the one thing you want when a download test fails in CI. Convention here is to thread a real logger through — and this file already has a `t`-backed one: `apiTestLogger` (lines 31-36) implements `Debugf`, so it already satisfies `httpfetch.Logger`.

But swapping the logger leaves the real oddity: a type named `fakeExecutableFetcher` constructing a live HTTP client inside one of its methods. The doc comment is honest about it, which is a good sign it wants to be expressed differently — this isn't a fake `Download` at all, it's a *real* `Download` plus a stubbed `DownloadWithHash`. Say that with composition:

```go
// fakeExecutableFetcher is a real httpfetch.Client with only DownloadWithHash
// stubbed: it records the requested executable download and writes a stub file at
// destPath, so the executable-activation path can be exercised without a
// hash-matching served artifact. Get and Download pass through to the embedded
// real client, which is what the step-source zip download needs.
type fakeExecutableFetcher struct {
	httpfetch.Client
	calledURL   string
	calledHash  string
	downloadErr error
}

func newFakeExecutableFetcher(t *testing.T) *fakeExecutableFetcher {
	return &fakeExecutableFetcher{Client: httpfetch.NewClient(apiTestLogger{t})}
}
```

Then `DownloadWithHash` stays exactly as it is, and both `Get` and `Download` are deleted. That gets you:

- `discardLogger` gone, real logger by construction.
- The real-I/O delegation visible at the callsite instead of hardcoded inside the fake.
- `Get` fixed as a side effect. It currently returns `(nil, nil)` — a nil reader with a nil error, which would nil-panic any caller that trusted it. Nothing calls it today (the tests build their own client via `steplibrary.New`), but it's a landmine, and pass-through removes it rather than leaving a lie in place.

There are 8 construction sites (`activate_executable_test.go:119,134`, `activate_test.go:32,52,67,81,107,133`), hence the constructor — with an embedded interface, a site that forgets to set it panics on first use, and `exhaustruct` won't catch it since `.golangci.yml:23-26` excludes `_test.go`. The one site that needs the failure mode becomes `f := newFakeExecutableFetcher(t); f.downloadErr = errFakeDownload`.

One alternative I'd argue *against*, for the record: making `Download` genuinely fake by having it write `writeStepSourceZip(f.t, destPath)` and ignore the URL. It would drop the HTTP and the logger question entirely, but it would also stop verifying that the download location was built correctly — `BuildStepSourceDownloadLocations` forming `<base><stepID>/<version>/step.zip`, which is exactly why `serveHelloStepInventoryWithExecutables` bothers to lay the archives out under `archive/hello-step/<version>/`. Better to keep that download real.

Scoped to the tests — not suggesting any change to `ActivateStep` taking a single shared fetcher.

pgyula:
PR description updated

stepman/util.go
pgyula: Agreed, and taken as recommended — DownloadStep now stages and renames, downloadStepZip's temp dir is unchanged.

stepman/util.go
pgyula: Right on both halves, and both are now spelled out in the description: the lost coverage (mid-body failures and corrupt-archive extraction were inside the old outer loop and are now outside any retry) and the win (4xx no longer burns 3 attempts and 6s before falling through to git).

stepman/util.go
pgyula: Fixed, and yes, log wanted threading through anyway — downloadStepZip(fetcher, url, destDir, log) now warns on a failed temp-dir cleanup.

activator/steplib/helpers_test.go
pgyula: Done exactly as suggested

internal/ziputil/ziputil_test.go
```diff
@@ -0,0 +1,110 @@
+package ziputil
```
lpusok: Pls use go-utils ziputil instead of copying it here

Badlazzor (approved):

PR #414, opened 2026-07-28
[STEP-2421] Report the StepLib inventory source in ActivatedStep

lpusok:
## Why

`ActivateSteplibRefStep` decides between the StepLib API and the legacy git-clone path (`shouldUseSteplibAPI`) and then throws that fact away: both branches return an identical `ActivatedStep`. The Bitrise CLI has no reference to `BITRISE_STEPLIB_API_ENABLE`, so it cannot infer the path itself — which makes every metric on its `cli_step_activation` event (duration, success rate) unattributable to a path. That attribution is what the gradual API rollout needs ([STEP-2421](https://bitrise.atlassian.net/browse/STEP-2421)).

## Next

Follow-ups from the STEP-2421 plan, each as its own PR: step source origin (`steplib_cache` / `zip` / `git`) with retry and fallback counts, and the precompiled outcome enum that distinguishes the four currently-silent fallbacks to source. The CLI-side telemetry property lands after a stepman release.

[STEP-2421]: https://bitrise.atlassian.net/browse/STEP-2421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/activator/result.go
+++ b/activator/result.go
@@ -42,8 +42,9 @@ type ActivationInventorySource string
 const (
 	// ActivationInventorySourceNone means the step was not activated from a StepLib (git or path ref).
 	ActivationInventorySourceNone ActivationInventorySource = ""
-	// ActivationInventorySourceSteplibAPI means the metadata came from the StepLib API, with no git clone.
+	// ActivationInventorySourceSteplibAPI means the metadata came from the StepLib API.
 	ActivationInventorySourceSteplibAPI ActivationInventorySource = "steplib_api"
-	// ActivationInventorySourceGitClone means the metadata came from the locally cloned StepLib and its spec.json.
-	ActivationInventorySourceGitClone ActivationInventorySource = "git_clone"
+	// ActivationInventorySourceSteplib means the metadata came from the locally set up StepLib
+	// and its generated spec.json.
+	ActivationInventorySourceSteplib ActivationInventorySource = "steplib"
 )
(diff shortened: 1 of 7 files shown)


activator/result.go
```diff
@@ -31,3 +35,15 @@ const (
...
+// ActivationInventorySource identifies which StepLib inventory served the step metadata.
+type ActivationInventorySource string
+
+const (
+	// ActivationInventorySourceNone means the step was not activated from a StepLib (git or path ref).
+	ActivationInventorySourceNone ActivationInventorySource = ""
+	// ActivationInventorySourceSteplibAPI means the metadata came from the StepLib API, with no git clone.
+	ActivationInventorySourceSteplibAPI ActivationInventorySource = "steplib_api"
+	// ActivationInventorySourceGitClone means the metadata came from the locally cloned StepLib and its spec.json.
+	ActivationInventorySourceGitClone ActivationInventorySource = "git_clone"
```
ofalvai: nit: technically, this is not always a git clone but a ZIP download. How about calling this `repo_download`?

lpusok: git clone is overloaded indeed, renaming.

lpusok: wdyt about just `steplib` ?

ofalvai (approved):

PR #415, opened 2026-07-30
Report the activation type on failed steplib activations

lpusok:
Follow-up to #414, from review of [bitrise#1297].

`ActivationType` was only assigned after the `steplib.ActivateStep` dispatch, but the legacy branch returns before it when `prepareStepLibForActivation` fails — and the caller keeps the partial result on error and reports it as telemetry.

So legacy activation failures reported `activation_type=""` while API failures reported a real type, since the API branch always reaches the dispatch. It is sent unconditionally, so that is an empty string in the warehouse, not an absent key: any dashboard grouping failures by activation type undercounts legacy failures to zero — an asymmetry in exactly the comparison #414 exists to enable.

Fix: set it to the source type up front. The dispatch still upgrades it on a successful precompiled download; a failure before that point never produced an executable, so source is accurate rather than a placeholder.

## Tests

Both tests returned right after `require.Error`, so nothing covered the partial result. They now assert it, and the assertions have teeth:

- revert the fix → all three legacy error cases fail with `expected: "steplib_source"`, `actual: ""`
- same revert → the API error case still passes, reproducing the asymmetry itself

`go build ./...`, `go vet ./...`, `gofmt -l activator/`, and the full `go test ./...` suite are clean.

[bitrise#1297]: https://github.com/bitrise-io/bitrise/pull/1297

### DIFF
--- a/activator/steplib_ref.go
+++ b/activator/steplib_ref.go
@@ -32,6 +32,9 @@ func ActivateSteplibRefStep(
 	activationResult := ActivatedStep{
 		StepYMLPath:      stepYMLPath,
 		DidStepLibUpdate: false,
+		// Set up front so the partial result kept on error still has a type; the
+		// dispatch below upgrades it on a successful precompiled download.
+		ActivationType: ActivationTypeSteplibSource,
 	}
 
 	var libraryAPI *steplibrary.Client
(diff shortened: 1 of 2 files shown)


lpusok:
Closing — not convinced the change is warranted.

For anyone finding this later: the behaviour it described is real (`ActivationType` is empty on legacy steplib activations that fail inside `prepareStepLibForActivation`, while API failures report a real type), but it only affects the failure slice of `cli_step_activation`, and we'd rather not add a pre-emptive default for it. The branch is left in place if it's worth revisiting.

PR #419, opened 2026-09-07
STEP-2520: build step activation through a constructor

lpusok:
[STEP-2520](https://bitrise.atlassian.net/browse/STEP-2520) · CLI side (also drops offline mode): https://github.com/bitrise-io/bitrise/pull/1357

Adds `activator.Activator`, built once per workflow run with `activator.New`, and makes `ActivateSteplibRefStep` a method on it. `ActivatePathRefStep` and `ActivateGitRefStep` stay package-level — they do no HTTP.

### One HTTP client per run

Every activation used to build **two** independent clients, each with its own transport and pool: one hard-wired inside `steplibrary.New`, one at the call site. No step reused a connection from the step before it, and the two clients in a single activation didn't share one either, despite both talking to `steplib.bitrise.io`.

The `Activator` now owns one `httpfetch.Client` and hands it to both. Measured against a local server, 10 sequential fetches:

| | connections |
|---|---|
| fresh client per call (before) | 10 |
| one shared client (after) | 1 |

`steplibrary.New` takes the fetcher as an argument instead of hard-wiring one — that's what makes the sharing possible.

### Config is supplied, not discovered

Stepman no longer reads any environment. Config is resolved once by the caller into `activator.Options`; the CLI does the mapping in `cli/step_activator.go`:

| env var (CLI-owned) | option |
|---|---|
| `BITRISE_STEPLIB_USE_API` | `Options.DisableSteplibAPI` |
| `BITRISE_STEPLIB_USE_BINARY` | `Options.DisablePrecompiled` |
| `BITRISE_STEPLIB_STORAGE_URLS` | `Options.PrecompiledStorageURLs` |

The flags are **opt-outs**, so the zero-value `Options` is what production wants (API + prebuilt executables on). Semantics are unchanged: unset leaves a feature on, `false`/`0` turns it off.

`steplib.ActivateStep` takes `steplib.Options` plus an explicit `useSteplibAPI bool`, instead of reading nil-ness of the `*steplibrary.Client` as "use the git-cloned steplib". The client is passed by value, so the flag and the client can't disagree; `steplibrary.Client` is now a documented copyable handle with value receivers.

### Why, beyond connection reuse

It gives the inventory client somewhere to hold state that outlives one activation — what an in-memory cache for the per-run-identical `step_ids.json` and `meta.json` needs. Today `step_ids.json` is the whole steplib's ID list, refetched per step to answer a `slices.Contains`. That cache is a follow-up ticket.

### Deliberately not here

- **`didStepLibUpdateInWorkflow` stays a parameter.** The CLI derives it per-steplib-source from `buildRunResults` and reads the result back; internalizing it needs a matching change there.
- **No `ctx` parameter.** `ActivateStep` still creates `context.Background()` internally. Natural follow-up, serves neither goal here.

### Breaking

Package-level `activator.ActivateSteplibRefStep` is gone, as is `OptionsFromEnv`. The Bitrise CLI is the only consumer and is migrated in #1357.

### Testing

`go build`, `go vet`, `golangci-lint` clean; `go test ./...` green, including new coverage for `useSteplibAPIFor`, `withDefaults`, the zero-value defaults and `steplib.Options.storageURLs()`. `_tests/activation` passes under `-tags integration`. Verified end to end through the CLI against the real steplib: unset takes the API path with no git clone and fetches the prebuilt binary; `USE_API=false` clones and makes zero API calls; `USE_BINARY=false` downloads no executable; `STORAGE_URLS` redirects the binary fetch at the given host.

[STEP-2520]: https://bitrise.atlassian.net/browse/STEP-2520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/activator/activator.go
+++ b/activator/activator.go
@@ -1,9 +1,6 @@
 package activator
 
 import (
-	"os"
-	"strings"
-
 	"github.com/bitrise-io/stepman/activator/steplib"
 	"github.com/bitrise-io/stepman/internal/httpfetch"
 	"github.com/bitrise-io/stepman/steplibrary"
@@ -13,14 +10,16 @@ import (
 const (
 	bitriseSteplibURL    = "https://github.com/bitrise-io/bitrise-steplib.git"
 	bitriseSteplibAPIURL = "https://steplib.bitrise.io/api"
-
-	useSteplibAPIEnv               = "BITRISE_STEPLIB_USE_API"
-	precompiledStepsEnv            = "BITRISE_STEPLIB_USE_BINARY"
-	precompiledStepsStorageURLsEnv = "BITRISE_STEPLIB_STORAGE_URLS"
 )
 
 // Options is the configuration of an Activator. It is resolved once, when the
 // Activator is built, rather than re-read on every step activation.
+//
+// Every field is supplied by the caller: stepman reads no configuration of its
+// own, so whichever process owns these settings also owns how they are
+// discovered. The Bitrise CLI, for one, maps environment variables onto them.
+// The zero value activates every step from source, through a git-cloned
+// StepLib.
 type Options struct {
 	// UseSteplibAPI activates steps of the canonical Bitrise StepLib over the
 	// StepLib V2 API instead of a local git clone. It has no effect on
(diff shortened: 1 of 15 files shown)


benbitrise:
Offline mode supports the air-gapped AWS product. Are we killing that product?

activator/activator.go
```diff
@@ -0,0 +1,122 @@
...
+	// IsOfflineMode forbids network access, restricting activation to what is
+	// already in the local StepLib cache. It is not supported together with
+	// UseSteplibAPI, which has no local inventory to read from.
+	IsOfflineMode bool
+}
+
+// OptionsFromEnv resolves the Options that stepman reads from the environment.
+// Both feature flags default to on and are disabled by setting them to "false"
+// or "0". Callers that configure the activator themselves do not need this.
+func OptionsFromEnv() Options {
```
benbitrise: Something to consider... do we even want stepman to read these env vars? Maybe responsiblity for that can entirely belong to bitrise cli, and stepman simply accepts whichever settings are provided to it.

benbitrise (dismissed):

activator/activator.go
lpusok: Agreed, done in bf42a8c.

benbitrise (approved):

PR #420, opened 2026-09-07
STEP-2525: remove StepLib offline mode

lpusok:
[STEP-2525](https://bitrise.atlassian.net/browse/STEP-2525)

> **Stacked on #419** (`STEP-2520-activator-constructor`). Review that one first — the diff shown here against `master` will include it until #419 merges.

Offline step activation is unused, and enabling the StepLib V2 API by default (#418) left it self-contradictory: the API path has no local inventory to read from, so an offline run against the canonical StepLib failed outright unless the user *also* opted out of the API with `BITRISE_STEPLIB_USE_API=false`. Rather than carry that through the activator refactor, drop the feature.

### Removed

- `IsOfflineMode` from `activator.Options` and `steplib.Options`
- the offline guard in `activateStepSourceWithAPI` (`activator/steplib/source.go`)
- the offline branch in `activateStepSource`, and `collectOfflineAvailableStepVersions`, which existed only to build its error message
- `canUpdateStepLib`, which existed only to combine the offline flag with `didStepLibUpdateInWorkflow` — `shouldUpdateStepLibForStep` now checks that directly

Net −110/+34.

### Kept on purpose

- **`ListCachedStepVersions`** — `cli/step_list.go` uses it independently of offline mode.
- **`preload/` and `bitrise preload-steps`** — warming the local StepLib cache is useful for build speed on its own. Only the CLI's help text referencing `BITRISE_OFFLINE_MODE` needs updating, which happens in the follow-up.

### Follow-up

The Bitrise CLI still defines `BITRISE_OFFLINE_MODE` and threads it down to this call. Removing that side is the second half of STEP-2525; after it, the env var stops having any effect. The CLI keeps compiling in the meantime — it just passes a flag that no longer reaches anything.

### Also in this PR

One stale doc pointer in `CLAUDE.md` that #419 introduced: it still referenced `shouldUseSteplibAPI` in `activator/steplib_ref.go`, which #419 replaced with `Activator.useSteplibAPIFor` in `activator/activator.go`. Fixed here rather than left to rot.

### Testing

- `go build ./...`, `go vet ./...`, `golangci-lint run` — clean
- `go test ./...` — green
- `_tests/activation` harness compiles under `-tags integration`; its offline sub-test and the `offline` parameter threaded through `activate()` / `logPair()` are gone

[STEP-2525]: https://bitrise.atlassian.net/browse/STEP-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


benbitrise:
Offline mode supports the air-gapped AWS product. Are we killing that product?

PR #421, opened 2026-09-09
STEP-2404: cache inventory reads in memory for the run

lpusok:
[STEP-2404](https://bitrise.atlassian.net/browse/STEP-2404)

Stepman was refetching the same StepLib inventory JSON (e.g. step_ids.json, 12.3KB) on every single step activation. This PR adds an in-memory HTTP cache for those inventory reads, cutting redundant requests.
The API publishes real cache headers, so the cache follows them instead of picking a TTL.

httpfetch.NewClients returns two clients that share one retryablehttp client, and therefore one connection pool.

```
retryablehttp.Client        (owns the pool)
  ├─ Passthrough  → uncached
  └─ Caching  → cacheSuccessesOnly → httpcache transport → same pool
```

[STEP-2404]: https://bitrise.atlassian.net/browse/STEP-2404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,16 +16,12 @@ Key entry point is `main.go` → `cli.Run()` which sets up the CLI application w
 
 ### Step activation paths
 
-Activation goes through `activator.Activator`, built once per run with `activator.New` (see `activator/activator.go`), so every step of a run shares one HTTP client. `Options` carries the config that used to be read from env vars deep in the call stack: stepman reads no environment of its own, and the caller — the Bitrise CLI — resolves those settings and passes them in. `activator/` has two ways to resolve and download a step:
+Activation goes through `activator.Activator`, built once per run with, so every step of a run shares one HTTP client.
 
 - **Legacy git-clone path**: clones the steplib repo locally and reads spec.json. This is the codepath used for custom/self-hosted steplibs.
-- **StepLib V2 API path** (default): fetches static JSON over HTTPS (no git clone), implemented in `activator/steplib/` (`activate.go`, `source.go`) and `steplibrary/`. Used for the canonical Bitrise steplib source when `Options.UseSteplibAPI` is set — see `Activator.useSteplibAPIFor` in `activator/activator.go`. The API base URL and the shared HTTP client are passed to `steplibrary.New`. Offline mode is not supported on this path and fails explicitly: an offline run against the canonical steplib has to opt out via `Options.UseSteplibAPI` to fall back to the git-clone path.
+- **StepLib API path** (default): fetches static JSON over HTTPS (no git clone).
 
-Precompiled step executables (skip building from source) are controlled by `Options.UsePrecompiled`, with the storage URLs to try in `Options.PrecompiledStorageURLs` (defaults in `activator/steplib/activate.go`).
-
-Inventory reads go through an in-memory HTTP cache (`github.com/bartventer/httpcache`) that honours the API's `Cache-Control` and revalidates with ETags, built in `httpfetch.NewClients` and living only for the run. Step archives and precompiled executables use the second, uncached client from the same pair; both share one connection pool.
-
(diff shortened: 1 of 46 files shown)


activator/activator.go
```diff
@@ -67,13 +67,24 @@ type Activator struct {
 // building one per step is what this constructor exists to avoid.
 func New(log stepman.Logger, opts Options) *Activator {
 	opts = withDefaults(opts)
-	fetcher := httpfetch.NewClient(log)
+
+	// The inventory client caches responses in memory; the download client
+	// does not. Both share one connection pool. The cache is an optimisation,
+	// so if it cannot be built the run continues without it rather than
+	// failing - but it says so, because a run silently paying full price for
+	// every inventory read is worth knowing about.
```
ofalvai: Why do we explain the implementation details of `httpfetch` here?

internal/httpfetch/httpfetch.go
```diff
@@ -16,9 +16,18 @@ import (
 	"os"
 	"path/filepath"
 
+	"log/slog"
+
+	"github.com/bartventer/httpcache"
+	// Registers the "memcache" store scheme used by inventoryCacheDSN.
```
ofalvai: Please clean up these unnecessary verbose LLM comments

internal/httpfetch/httpfetch.go
```diff
@@ -16,9 +16,18 @@ import (
...
+	_ "github.com/bartventer/httpcache/store/memcache"
 	"github.com/hashicorp/go-retryablehttp"
 )
 
+// inventoryCacheDSN selects the library's in-memory store. Nothing is written
+// to disk: the cache lives and dies with the process.
+const inventoryCacheDSN = "memcache://"
```
ofalvai: Does this matter this much? Why do we need to explain this and extract to a constant? I'd just inline this into `httpcache.NewTransport()`

internal/httpfetch/httpfetch.go
```diff
@@ -62,6 +95,104 @@ func NewClient(logger Logger) Client {
...
+		Downloads: &client{httpClient: downloads},
+	}, nil
+}
+
+// newCachingTransport wraps upstream with the in-memory response cache.
+// httpcache.NewTransport panics with ErrOpenCache rather than returning an
+// error when the store cannot be opened, so turn that one back into an error:
+// a cache that is only an optimisation must not take down a build. Anything
+// else is a genuine bug and keeps its stack.
+func newCachingTransport(logger Logger, upstream http.RoundTripper) (rt http.RoundTripper, err error) {
```
ofalvai: nitpick, but some of these cache-related functions/types, as well as the logging adapter could be in separate files within this same package. This file is getting a bit long.

ofalvai (changes requested):
LGTM functionally, but please clean up the overly verbose LLM comments

activator/activator.go
ofalvai: This is no longer true. This is why I don't like these kind of overly verbose comments.

CLAUDE.md
```diff
@@ -23,6 +23,8 @@ Activation goes through `activator.Activator`, built once per run with `activato
 
 Precompiled step executables (skip building from source) are controlled by `Options.UsePrecompiled`, with the storage URLs to try in `Options.PrecompiledStorageURLs` (defaults in `activator/steplib/activate.go`).
 
+Inventory reads go through an in-memory HTTP cache (`github.com/bartventer/httpcache`) that honours the API's `Cache-Control` and revalidates with ETags, built in `httpfetch.NewClients` and living only for the run. Step archives and precompiled executables use the second, uncached client from the same pair; both share one connection pool.
```
ofalvai: I wouldn't include these in the project's top-level `CLAUDE.md`. This information is rarely needed, bloats the context window for every session, and the agent can just go and read the source if it ever needs to figure it out.

ofalvai (approved):

